PR #1, opened 2022-04-25
WIP: Add CRD for OSBuildConfig

ygalblum:
This is the initial CRD for the OSBuildConfig. This PR is split to three commits out of which the first and third are generated code and only the second one includes code that I have added.

The first commit is the result of running:
`operator-sdk create api --version v1alpha1 --kind OSBuildConfig --resource --controller`

The third commit is the result of running `make && make manifests` after the second commit


api/v1alpha1/osbuildconfig_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// Parent is the ref of the parent of target build (Optional)
+	Parent *string `json:"parent"`
+}
+
+// +kubebuilder:validation:Enum=edge-container;edge-installer
+type ImageType string
+
+// BuildDetails includes all the information needed to build the image
+type BuildDetails struct {
+	// Distribution is the name of the O/S distribution
```
bardielle, line 51: I suggested that all the mandatory fields will appear before the optional

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// EnabledServices is the list of services to enabled (optional)
+	EnabledServices []string `json:"enabled_services,omitempty"`
+	// DisabledServices is the list of services to disabled (optional)
+	DisabledServices []string `json:"disabled_services,omitempty"`
+	// Architecture defines target architecture of the image
+	Architecture string `json:"architecture"`
+	// OSTree is the OSTree configuration of the build
+	OSTree *OSTreeConfig `json:"ostree,omitempty"`
+	// ImageType defines the target image type
+	ImageType ImageType `json:"image_type"`
```
bardielle, line 66: ImageType->TargetImageType

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// Distribution is the name of the O/S distribution
+	Distribution string `json:"distribution"`
+	// Packages is a list of RPM packages to install (optional)
+	Packages []string `json:"packages,omitempty"`
+	// Users is the list of Users to add to the image (optional)
+	Users []User `json:"users,omitempty"`
+	// EnabledServices is the list of services to enabled (optional)
+	EnabledServices []string `json:"enabled_services,omitempty"`
+	// DisabledServices is the list of services to disabled (optional)
+	DisabledServices []string `json:"disabled_services,omitempty"`
```
bardielle, line 60: Maybe all the kickstart file sections should be on a different struct?

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+type BuildTriggers struct {
+	// ConfigChange if True trigger a new build upon any change in this BuildConfig CR
+	ConfigChange *bool `json:"config_change,omitempty"`
+}
+
+// OSBuildConfigSpec defines the desired state of OSBuildConfig
+type OSBuildConfigSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
```
bardielle, line 76: Can you add an example.yaml file of that CR?

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -0,0 +1,116 @@
...
+	Architecture string `json:"architecture"`
+	// OSTree is the OSTree configuration of the build
+	OSTree *OSTreeConfig `json:"ostree,omitempty"`
+	// ImageType defines the target image type
+	ImageType ImageType `json:"image_type"`
+}
+
+type BuildTriggers struct {
+	// ConfigChange if True trigger a new build upon any change in this BuildConfig CR
+	ConfigChange *bool `json:"config_change,omitempty"`
```
bardielle, line 71: What about the webhook?

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Details defines what to build
+	Details BuildDetails `json:"build_details"`
+	// Triggers defines when to build
+	Triggers BuildTriggers `json:"build_triggers"`
+}
+
+// OSBuildConfigStatus defines the observed state of OSBuildConfig
+type OSBuildConfigStatus struct {
```
bardielle, line 86: what about the triggers so they can be tracked?

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// Important: Run "make" to regenerate code after modifying this file
+
+	LastVersion *int `json:"lastversion,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
+// OSBuildConfig is the Schema for the osbuildconfigs API
+type OSBuildConfig struct {
```
bardielle, line 97: Can you add an example.yaml file of that CR? 

Also I think that this should be appear first in this file (I mean start from the main struct and explode it )

config/samples/_v1alpha1_osbuildconfig.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: osbuilder.project-flotta.io/v1alpha1
+kind: OSBuildConfig
+metadata:
+  name: osbuildconfig-sample
+spec:
+  # TODO(user): Add fields here
```
bardielle, line 6: Please fill here the sample as part of the PR

api/v1alpha1/osbuildconfig_types.go
ygalblum: These do not belong to the kickstart. Blueprint files already support this feature, while the REST V2 API do not. We have a task for it adding this capability

api/v1alpha1/osbuildconfig_types.go
ygalblum: Since we cannot use the OpenShift API server (as the BuildConfig does), I wasn't sure how to handle it and which information will be needed. But, I think we can have something similar to that, which will also just require a secret.

api/v1alpha1/osbuildconfig_types.go
ygalblum: Not sure I agree with this approach. I think that fields that relate to each other should be together even if some are mandatory and some are optional

api/v1alpha1/osbuildconfig_types.go
ygalblum: Are you referring to the type, name or both?

bardielle: both

api/v1alpha1/osbuildconfig_types.go
bardielle: I referred to the build....

ygalblum:
Closing the PR after changing the tooling to kubebuilder

PR #8, opened 2022-05-15
Validation admission webhook

bardielle:
This PR depends on [that](https://github.com/project-flotta/osbuild-operator/pull/3) PR. 

As part of the OSBuildConfig CRD, some fields should be defined as immutable - Architecture, Distributing,  and TargetImageType. 
Those fields define something basic that changing might change the consequent builds and not linked to the first OSBuild.


api/v1alpha1/osbuildconfig_webhook.go
```diff
@@ -0,0 +1,91 @@
...
+	if r.Spec.Details.Distribution != oldOSBuildConfig.Spec.Details.Distribution {
+		osbuildconfiglog.Error(err, "Distribution is an immutable field and cannot be updated")
+		return err
+	}
+
+	if r.Spec.Details.TargetImage.TargetImageType != oldOSBuildConfig.Spec.Details.TargetImage.TargetImageType {
+		osbuildconfiglog.Error(err, "TargetImageType is an immutable field and cannot be updated")
+		return err
+	}
+
```
ygalblum, line 81: TargetImage.OSTree is also immutable

ygalblum (commented):
Next time, for the review, please split into two commits the auto-generate code and yours so that your changes can be easily seen

api/v1alpha1/osbuildconfig_webhook.go
bardielle: Done, 
Also, I split it into 3 different commits

api/v1alpha1/osbuildconfig_webhook.go
```diff
@@ -0,0 +1,98 @@
...
+	if r.Spec.Details.TargetImage.OSTree != oldOSBuildConfig.Spec.Details.TargetImage.OSTree {
+		osbuildconfiglog.Error(err, "OSTree is an immutable field and cannot be updated")
```
masayag, line 85: shouldn't use deep copy for OSTree comparison?

config/default/manager_webhook_patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+  template:
+    spec:
+      containers:
+      - name: manager
+        ports:
+        - containerPort: 9443
+          name: webhook-server
+          protocol: TCP
+        volumeMounts:
+        - mountPath: /tmp/k8s-webhook-server/serving-certs
```
masayag, line 16: attempt to run the PR locally with `make run` and on the cluster by `make deploy` fails with:
```
E0518 04:52:31.676884       1 leaderelection.go:330] error retrieving resource lock osbuild-operator-system/bfdcaedc.osbuilder.project-flotta.io: Get "https://10.96.0.1:443/apis/coordination.k8s.io/v1/namespaces/osbuild-operator-system/leases/bfdcaedc.osbuilder.project-flotta.io": context canceled
1.652849551676954e+09   ERROR   setup   problem running manager {"error": "open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory"}
main.main
        /workspace/main.go:105
runtime.main
        /usr/local/go/src/runtime/proc.go:255
```

main.go
```diff
@@ -85,6 +85,10 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "OSBuildConfig")
 		os.Exit(1)
 	}
+	if err = (&osbuilderprojectflottaiov1alpha1.OSBuildConfig{}).SetupWebhookWithManager(mgr); err != nil {
```
masayag, line 88: in flotta for simplify dev-env, webhooks are disabled by default when running `make run` by exposing env-var.
see https://github.com/project-flotta/flotta-operator/blob/main/main.go#L132

api/v1alpha1/osbuildconfig_webhook.go
```diff
@@ -0,0 +1,107 @@
...
+		osbuildconfiglog.Error(err, "Distribution is an immutable field and cannot be updated")
+		return err
+	}
+
+	if r.Spec.Details.TargetImage.TargetImageType != oldOSBuildConfig.Spec.Details.TargetImage.TargetImageType {
+		osbuildconfiglog.Error(err, "TargetImageType is an immutable field and cannot be updated")
+		return err
+	}
+
+	if r.Spec.Details.TargetImage.OSTree.URL != oldOSBuildConfig.Spec.Details.TargetImage.OSTree.URL {
```
ygalblum, line 84: Since all of these are pointers, you need to compare their contents

config/default/manager_webhook_patch.yaml
ygalblum: @masayag Is this issue specific to this PR? Sounds like it's from the previous commit

ygalblum (changes requested):

config/default/manager_webhook_patch.yaml
masayag: yes, the error includes a missing file which is referred to by this PR.
I don't get that error without it.

masayag: after taking the update PR, the error isn't reproduced.

ygalblum (approved):

PR #11, opened 2022-05-18
Add sample content and group name to osbuild-config sample 

masayag:
Signed-off-by: Moti Asayag <masayag@redhat.com>

### DIFF
--- a/config/samples/osbuilder_v1alpha1_osbuildconfig.yaml
+++ b/config/samples/osbuilder_v1alpha1_osbuildconfig.yaml
@@ -1,0 +1,19 @@
+apiVersion: osbuilder.project-flotta.io/v1alpha1
+kind: OSBuildConfig
+metadata:
+  name: osbuildconfig-sample
+spec:
+  details:
+    distribution: fedora-35
+    customizations:
+      packages:
+        - node_exporter
+        - ansible
+    targetImage:
+      architecture: x86_64
+      targetImageType: edge-container
+      osTree:
+        url: "https://hostedimagesserver.flotta.io/"
+        ref: "rhel/8/x86_64/edge"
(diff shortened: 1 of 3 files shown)


config/samples/osbuilder_v1alpha1_osbuildconfig.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: osbuilder.project-flotta.io/v1alpha1
+kind: OSBuildConfig
+metadata:
+  name: osbuildconfig-sample
+spec:
+  details:
+    distribution: Fedora
```
ygalblum, line 7: Distribution is lower case and includes the version. E.g. `fedora-35`

ygalblum (changes requested):

config/samples/osbuilder_v1alpha1_osbuildconfig.yaml
masayag: should there be a validation for that field structure in the CRD?
what would be the desired pattern for it? we can add on top of https://github.com/project-flotta/osbuild-operator/blob/main/api/v1alpha1/osbuildconfig_types.go#L41 something like:
```
// +kubebuilder:validation:Pattern=^[a-z0-9-.]+$
```

is this field validate on the composer API side and has a known format to it?

ygalblum: There is no validation for it on the Composer side either, nor is there in the Weldr API (or defined in the Blueprint document https://www.osbuild.org/guides/blueprint-reference/blueprint-reference.html#distribution-selection-with-blueprints)

ygalblum (approved):

PR #13, opened 2022-05-18
Add osbuild crd

bardielle:

PR #15, opened 2022-05-18
OSBuildConfig controller

bardielle:
After creating a new OSBuildConfig CR or updating an existing CR (by the user) a new OSBuild CR should be created.
The OSBuildConfig controller is triggering a new OSBuild CR per update or creation of OSBuildConfig CR

Signed-off-by: Danielle Barda [dbarda@redhat.com](mailto:dbarda@redhat.com)

### DIFF
--- a/controllers/osbuildconfig_controller.go
+++ b/controllers/osbuildconfig_controller.go
@@ -18,19 +18,25 @@ package controllers
 
 import (
 	"context"
-
+	"github.com/go-logr/logr"
+	osbuilderprojectflottaiov1alpha1 "github.com/project-flotta/osbuild-operator/api/v1alpha1"
+	"github.com/project-flotta/osbuild-operator/internal/repository/osbuild"
+	"github.com/project-flotta/osbuild-operator/internal/repository/osbuildconfig"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
-
-	osbuilderprojectflottaiov1alpha1 "github.com/project-flotta/osbuild-operator/api/v1alpha1"
 )
 
 // OSBuildConfigReconciler reconciles a OSBuildConfig object
 type OSBuildConfigReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme                  *runtime.Scheme
+	OSBuildConfigRepository osbuildconfig.Repository
+	OSBuildRepository       osbuild.Repository
 }
 
 //+kubebuilder:rbac:groups=osbuilder.project-flotta.io,resources=osbuildconfigs,verbs=get;list;watch;create;update;patch;delete
@@ -47,11 +53,70 @@ type OSBuildConfigReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/reconcile
 func (r *OSBuildConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx)
+	logger.Info("Reconciling", "OSBuildConfig", req)
 
-	// TODO(user): your logic here
+	osBuildConfig, err := r.OSBuildConfigRepository.Read(ctx, req.Name, req.Namespace)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{Requeue: true}, err
+	}
+
+	if osBuildConfig.DeletionTimestamp != nil {
+		// The OSBuild CRs that were created by that OSBuildConfig would be deleted
+		// thanks to setting controller reference for each OSBuild CR
+		return ctrl.Result{}, nil
+	}
+
+	err = r.createNewOSBuildCR(ctx, osBuildConfig, logger)
+	if err != nil {
+		return ctrl.Result{Requeue: true}, err
+	}
 
 	return ctrl.Result{}, nil
+}
+
+func (r *OSBuildConfigReconciler) createNewOSBuildCR(ctx context.Context, osBuildConfig *osbuilderprojectflottaiov1alpha1.OSBuildConfig, logger logr.Logger) error {
+	osBuildNewVersion := *osBuildConfig.Status.LastVersion + 1
+
+	osBuildConfigSpecDetails := osBuildConfig.Spec.Details.DeepCopy()
+	osBuild := &osbuilderprojectflottaiov1alpha1.OSBuild{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: osBuildConfig.Name + "-" + string(rune(osBuildNewVersion)),
+		},
+		Spec: osbuilderprojectflottaiov1alpha1.OSBuildSpec{
+			Details:     *osBuildConfigSpecDetails,
+			TriggeredBy: "UpdateCR",
+		},
+	}
+
+	// Set the owner of the osBuild CR to be osBuildConfig in order to manage lifecycle of the osBuild CR.
+	// Especially in deletion of osBuildConfig CR
+	err := controllerutil.SetControllerReference(osBuildConfig, osBuild, r.Scheme)
+	if err != nil {
+		logger.Error(err, "cannot create osBuild")
+		return err
+	}
+
+	patch := client.MergeFrom(osBuildConfig.DeepCopy())
+	osBuildConfig.Status.LastVersion = &osBuildNewVersion
+	err = r.OSBuildConfigRepository.PatchStatus(ctx, osBuildConfig, &patch)
+	if err != nil {
+		logger.Error(err, "cannot update the field lastVersion of osBuildConfig")
+		return err
+	}
+
+	err = r.OSBuildRepository.Create(ctx, osBuild)
+	if err != nil {
+		logger.Error(err, "cannot create osBuild")
+		return err
+	}
+
+	logger.Info("A new OSBuild CR was created", osBuild.Name)
+
+	return nil
 }
 
 // SetupWithManager sets up the controller with the Manager.
--- a/internal/repository/osbuild/osbuild.go
+++ b/internal/repository/osbuild/osbuild.go
@@ -1,0 +1,31 @@
+package osbuild
+
+import (
+	"context"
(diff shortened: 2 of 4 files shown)


controllers/osbuildconfig_controller.go
```diff
@@ -47,13 +53,54 @@ type OSBuildConfigReconciler struct {
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{Requeue: true}, err
+	}
 
-	// TODO(user): your logic here
+	if osBuildConfig.DeletionTimestamp != nil {
+		return ctrl.Result{}, nil
```
ygalblum, line 68: I think we should delete all OSBuild CRs linked to this OSBuildConfig when the latter is deleted

controllers/osbuildconfig_controller.go
```diff
@@ -47,13 +53,54 @@ type OSBuildConfigReconciler struct {
...
+		return ctrl.Result{Requeue: true}, err
+	}
 
 	return ctrl.Result{}, nil
 }
 
+func (r *OSBuildConfigReconciler) createOSBuildCR(ctx context.Context, osBuildConfig *osbuilderprojectflottaiov1alpha1.OSBuildConfig, logger logr.Logger) error {
+	osBuildCurrentVersion := string(rune(*osBuildConfig.Status.LastVersion + 1))
+
+	osBuildConfigCp := osBuildConfig.DeepCopy()
```
ygalblum, line 82: Do you really need to make a copy of the OSBuildConfig structure? If you do, then maybe just its `Spec.Details`

controllers/osbuildconfig_controller.go
```diff
@@ -47,13 +53,54 @@ type OSBuildConfigReconciler struct {
...
+		},
+	}
+
+	err := r.OSBuildRepository.Create(ctx, osBuild)
+	if err != nil {
+		logger.Error(err, "cannot create osBuild")
+		return err
+	}
+	logger.Info("A new OSBuild CR was created", osBuild.Name)
+
```
ygalblum, line 100: I think we need to update the LastVersion field of the OSBuildConfig, no?

controllers/osbuildconfig_controller.go
```diff
@@ -47,13 +53,54 @@ type OSBuildConfigReconciler struct {
...
+	}
 
 	return ctrl.Result{}, nil
 }
 
+func (r *OSBuildConfigReconciler) createOSBuildCR(ctx context.Context, osBuildConfig *osbuilderprojectflottaiov1alpha1.OSBuildConfig, logger logr.Logger) error {
+	osBuildCurrentVersion := string(rune(*osBuildConfig.Status.LastVersion + 1))
+
+	osBuildConfigCp := osBuildConfig.DeepCopy()
+	// copy to osbuildCR if not exist
```
ygalblum, line 83: It's a comment, but, what do you mean by `if not exist`?

controllers/osbuildconfig_controller.go
```diff
@@ -30,7 +34,9 @@ import (
 // OSBuildConfigReconciler reconciles a OSBuildConfig object
 type OSBuildConfigReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme                  *runtime.Scheme
+	OSBuildConfigRepository osbuildconfig.Repository
```
ygalblum, line 38: Don't you need to initialize these two with a client?

ygalblum:
Is this PR supposed to be on top of #13 ?

controllers/osbuildconfig_controller.go
bardielle: AFAIN it's happening automatically

controllers/osbuildconfig_controller.go
bardielle: Fixed it

controllers/osbuildconfig_controller.go
bardielle: removed that comment

controllers/osbuildconfig_controller.go
bardielle: You're right, thanks
Fixed it

controllers/osbuildconfig_controller.go
ygalblum: Sorry, this was supposed to be put here:
AFAIK you need to set the owner for that to happen.
See example usage here: https://github.com/ctera/kubefiler-operator/blob/2a06c8bd0f0b9ba04dd783c8d6c4867762fc4f87/internal/resources/deployment.go#L88
See documentation here: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetControllerReference

controllers/osbuildconfig_controller.go
ygalblum: Sorry, disregard my previous comment. It won't happen automatically. The controller needs to update it. Something like this: https://github.com/ctera/kubefiler-operator/blob/2a06c8bd0f0b9ba04dd783c8d6c4867762fc4f87/internal/resources/kubefiler.go#L186

controllers/osbuildconfig_controller.go
ygalblum: The OSBuild controller will be notified on the deleting of the CR. It will then need to understand what to do

api/v1alpha1/osbuild_types.go
```diff
@@ -25,19 +25,45 @@ import (
...
 type OSBuildSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// Details defines what to build
+	Details BuildDetails `json:"details"`
 
-	// Foo is an example field of OSBuild. Edit osbuild_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// TriggeredBy explains what triggered the build out
+	TriggeredBy TriggeredBy `json:"config"`
```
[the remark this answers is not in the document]
bardielle: You're right. It should be part of this PR: https://github.com/project-flotta/osbuild-operator/pull/13

api/v1alpha1/osbuild_types.go
```diff
@@ -25,19 +25,45 @@ import (
...
 }
 
+// +kubebuilder:validation:Enum=UpdateCR;Webhook
+type TriggeredBy string
+
```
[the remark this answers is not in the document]
bardielle: Right now we are supposed to support only 2 types - rest API(webhook) and update the CR and trigger the reconcile 
If you think about more options please share and I will add it to the enum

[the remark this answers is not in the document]
jordigilh: None at the moment. Just wondering if you were going to create the instances later on or still considering which ones. Here it would be a good opportunity to make use of the types rather than a hardcoded string, what do you think?
https://github.com/project-flotta/osbuild-operator/blob/1baa26ea81d7db314457645dd86cdb74b112e4d0/controllers/osbuildconfig_controller.go#L88-L91

config/samples/osbuilder_v1alpha1_osbuild.yaml
```diff
@@ -3,4 +3,16 @@ kind: OSBuild
 metadata:
   name: osbuild-sample
 spec:
-  # TODO(user): Add fields here
+  details:
+    distribution: fedora-35
```
jordigilh, line 7: s/fedora-35/rhel-90/

Fedora does not support `edge-container` image type, just `iot-container`.

controllers/osbuildconfig_controller.go
```diff
@@ -47,13 +53,64 @@ type OSBuildConfigReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/reconcile
 func (r *OSBuildConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx)
+	logger.Info("Reconciling", "edgeDevice", req)
```
jakub-dzon, line 57: edgeDevice->osBuildConfig

internal/repository/osbuild/osbuild.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"context"
+	"github.com/project-flotta/osbuild-operator/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//go:generate mockgen -package=osbuild -destination=mock_osbuild.go . Repository
+type Repository interface {
```
jakub-dzon, line 11: Are all of the methods needed? I would keep it lean and add new ones when needed.

internal/repository/osbuildconfig/osbuildconfig.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"context"
+	"github.com/project-flotta/osbuild-operator/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//go:generate mockgen -package=osbuildconfig -destination=mock_osbuildconfig.go . Repository
+type Repository interface {
+	Read(ctx context.Context, name string, namespace string) (*v1alpha1.OSBuildConfig, error)
```
jakub-dzon, line 11: Same here - I don't think all of the methods are needed at this point. Please leave only used ones.

jakub-dzon (changes requested):
Please rebase - #13 has been merged.

internal/repository/osbuildconfig/osbuildconfig.go
```diff
@@ -0,0 +1,31 @@
...
+	Create(ctx context.Context, osBuildConfig *v1alpha1.OSBuildConfig) error
```
jakub-dzon, line 12: This one is not used as well - is there a reason for leaving it here?

controllers/osbuildconfig_controller.go
ygalblum: This only updated the local structure, not the object stored in etcd. You need to run `client.Status().Update(ctx, osBuildConfig)`.  See the next line: https://github.com/ctera/kubefiler-operator/blob/2a06c8bd0f0b9ba04dd783c8d6c4867762fc4f87/internal/resources/kubefiler.go#L187

ygalblum (changes requested):
See my comment about lastVersion

bardielle:
@ygalblum I'm updating lastVersion here : https://github.com/project-flotta/osbuild-operator/pull/15/files#diff-9c1ccd67b2d4d3b3ab036719de0f6b541dffa0a17a207bd58d0064b820e36155R109
Do you think it's not enough?

ygalblum:
> @ygalblum I'm updating lastVersion here : https://github.com/project-flotta/osbuild-operator/pull/15/files#diff-9c1ccd67b2d4d3b3ab036719de0f6b541dffa0a17a207bd58d0064b820e36155R109 Do you think it's not enough?

See my last comment. This line only updated the local Golan struct. It does not save it on the server

jakub-dzon (approved):

ygalblum (approved):

PR #19, opened 2022-05-25
Add basic CI tests to GitHub workflows

ygalblum:
Make sure breaking code does not get merged
Since there are no unit tests yet, set coverage fail at 0

Fix lint error

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,62 @@
+---
+name: Go
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  Build:
+    runs-on: ubuntu-latest
+    outputs:
+      coverage: ${{ steps.coverage.outputs.coverage }}
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17
+
+    - name: Build
+      run: make build
+
+    - name: Test
+      run: make test
+
+    - name: Coverage
+      id: coverage
+      run: |
+        make test-create-coverage
+        COVERAGE=$(go tool cover --func=cover.out |  grep total | grep -Eo '[0-9]+\.[0-9]+')
+        echo "::set-output name=coverage::${COVERAGE}"
+
+    - uses: actions/upload-artifact@v2
+      with:
+        name: test-coverage
+        path: coverage.html
+
+    - name: Checking generated files are up to date
+      run: |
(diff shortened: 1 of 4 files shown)


api/v1alpha1/webhook_suite_test.go
```diff
@@ -31,7 +31,7 @@ import (
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	//+kubebuilder:scaffold:imports
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/rest"
+	// "k8s.io/client-go/rest"
```
bardielle, line 34: can it be removed?

api/v1alpha1/webhook_suite_test.go
```diff
@@ -43,7 +43,7 @@ import (
 // These tests use Ginkgo (BDD-style Go testing framework). Refer to
 // http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
 
-var cfg *rest.Config
+// var cfg *rest.Config
```
bardielle, line 46: Can it be removed?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+      coverage: ${{ steps.coverage.outputs.coverage }}
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17
+
+    - name: Test
```
bardielle, line 22: what do you think about adding make build as a workflow/job?

controllers/suite_test.go
```diff
@@ -23,7 +23,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/rest"
+	// "k8s.io/client-go/rest"
```
bardielle, line 26: Also here - can you remove those lines?

api/v1alpha1/webhook_suite_test.go
ygalblum: I'll reply once for all 4 commented lines. I commented instead of deleting them because this is generated code, so I assume we will need it once we implement the test. But, if you think it's better to remove them, I don't mind

.github/workflows/ci.yml
ygalblum: I thought `make test` covered it, but it seems that it doesn't really do that. So, sure, I've added it now

api/v1alpha1/webhook_suite_test.go
bardielle: So let's remove it and maybe open a relevant task in Jira for implementing those tests

ygalblum: Done

PR #29, opened 2022-05-31
Add OSBuildConfigTemplate CRD

jakub-dzon:


ygalblum:
@jakub-dzon in last week's call we decided to enforce that the branch will be updated with the base branch. So, you'll need to rebase before merging

jakub-dzon:
@ygalblum done.

PR #33, opened 2022-06-02
Implement the CRD for the OSBuildEnvConfig

ygalblum:
The CRD I have aligns more with the parameters required for the Helm deployment (https://github.com/ygalblum/osbuild-composer-openshift-helm) that with the design document since the latter was written before I fully understood what is required.

The CRD will need to change in a couple of places to allow the user different levels of control (for example, setting the server URL for the subscription manager). But, I believe this is a good place to start.

Please note that the second commit is the actual code I've added. The first and second are generated code and the fourth is the sample


api/v1alpha1/osbuildenvconfig_types.go
```diff
@@ -25,11 +26,126 @@ import (
 
 // OSBuildEnvConfigSpec defines the desired state of OSBuildEnvConfig
 type OSBuildEnvConfigSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// Composer contains all the required configuration values for the Composer service
```
bardielle, line 29: Do you think we need to limit the version of the osbuild-composer?

ygalblum: Not sure what you mean? The image tag and the OSBuild-Worker and OSBuild versions? If yes, then I don't think we want the user to control that at the CR level. It will create a support matrix from hell

bardielle: Lets talk about it in the sync meeting and open a relevant ticket if needed

bardielle (approved):

api/v1alpha1/osbuildenvconfig_types.go
ygalblum: Resolving this conversation. If we decide to allow the user to set the version, it will be added in a separate PR

PR #34, opened 2022-06-02
Add kickstart

jakub-dzon:
Adding support for providing Kickstart file for `OSBuild`


bardielle (approved):

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -43,6 +43,8 @@ type BuildDetails struct {
 	Customizations *Customizations `json:"customizations,omitempty"`
 	// TargetImage defines the requested output image
 	TargetImage TargetImage `json:"targetImage"`
+	// Kickstart may store content of a kickstart file to be used in the target image
+	Kickstart *string `json:"kickstart,omitempty"`
```
ygalblum, line 47: We wanted to restrict the `Kickstart` field to the `OSBuildConfigTemplate` CRD and not have it in the `OSBuildConfig` (otherwise we'll need also to handle text merge). So, this field should only be added to the `OSBuild` CRD

ygalblum (changes requested):

api/v1alpha1/osbuildconfig_types.go
jakub-dzon: Done.

bardielle: @ygalblum  Don't we want a default kickstart file in the osbuilcConfig? in case it wasn't created by the template?

bardielle: So we need to restrict editing OSBuildConfig that was created by the template so the template cannot be changed (webhook validator)

ygalblum (approved):

PR #35, opened 2022-06-03
Add template ref

jakub-dzon:
This PR adds support for referencing `OSBuildConfigTemplate` from `OSBuildConfigs`.  

This change also lays groundwork for ability to reconcile `OSBuildConfig` when `OSBuildConfigTemplate` changes - the new status fields. The algorithm would be as follow:
 - user adds a template reference to the `OSBuildConfig`
 - `OSBuildConfig.Status.CurrentTemplateResourceVersion` and `OSBuildConfig.Status.LastTemplateResourceVersion` are set to the current value of `OSBuildConfigTemplate.ObjectMeta.ResourceVersion`;
 - `OSBuild` is created;
 -  ...
 - User updates `OSBuildConfigTemplate`;
 - `OSBuildConfigTemplate` controller retrieves all `OSBuildConfig` objects using reconciled `OSBuildConfigTemplate` and sets their `Status.CurrentTemplateResourceVersion` - based on custom index;
 - preceding step triggers `OSBuildConfig` reconciliations which create new `OSBuild` resources


ygalblum:
I don't understand what the TemplateResourceVersion fields are used for. The snapshot of the build configuration is saved in the OSBuild CR.

ygalblum (changes requested):
Setting "Request Changes" since regular comments do not have a "resolve" button

jakub-dzon:
> I don't understand what the TemplateResourceVersion fields are used for. The snapshot of the build configuration is saved in the OSBuild CR.

`CurrentTemplateResourceVersion` update is supposed to be used to trigger new `OSBuildConfig` reconciliation after `OSBuildConfigTemplate` changes and `LastTemplateResourceVersion` is just for information purposes. We need a way to execute `OSBuildConfig` reconciliation logic (`OSBuild` creation) somehow.

ygalblum:
> > I don't understand what the TemplateResourceVersion fields are used for. The snapshot of the build configuration is saved in the OSBuild CR.
> 
> `CurrentTemplateResourceVersion` update is supposed to be used to trigger new `OSBuildConfig` reconciliation after `OSBuildConfigTemplate` changes and `LastTemplateResourceVersion` is just for information purposes. We need a way to execute `OSBuildConfig` reconciliation logic (`OSBuild` creation) somehow.

I see. So it comes instead of creating an OSBuild CR from the OSBuildConfigTemplate controller?
If so, then I think `LastTemplateResourceVersion` is not only informative, it allows the controller to recognize that the trigger for the reconciliation is the change in the  template and then check for the build trigger flag. Right?

jakub-dzon:
> I see. So it comes instead of creating an OSBuild CR from the OSBuildConfigTemplate controller? If so, then I think `LastTemplateResourceVersion` is not only informative, it allows the controller to recognize that the trigger for the reconciliation is the change in the template and then check for the build trigger flag. Right?

Exactly. I wanted to have only one reconciler to be responsible for creating `OSBuild` CRs. 
Come to think of it - yes, `LastTemplateResourceVersion` comparison with `CurrentTemplateResourceVersion` can be used to detect the source of change.

jakub-dzon:
> On one hand I agree with your point of having one reconciler that creates the CRs. Having said that the webhook trigger creates these CRs as well. @bardielle created a separate interface for it so that different actors can use. What do you think?

I see following problem with creating OSBuild CR in the template reconciler - in one reconcile loop we need to get all OSBuildConfig objects and for each create OSBuild CR; if something goes wrong, we need to redo the whole loop over all OSBuildConfig resources (ignoring already processed). When we leave that to the OSBuildConfig reconciler, the objects would be processed separately and if one has issues, only that one would be requeued.

ygalblum:
> in one reconcile loop we need to get all OSBuildConfig objects and for each create OSBuild CR
Yes, but you still have to go over all the OSBuildConfig objects and update their status. Having said that, I think it is faster, simpler and easier to implement in an idempotent way that creating the OSBuild CRs.
On the webhook side, it creates only one CR so it is simpler.

Ok, I'm convinced :)

ygalblum:
I've approved the PR and enabled auto-merge. Just rebase on top of main

PR #36, opened 2022-06-03
[WIP] Use OSBuildConfig template

jakub-dzon:


internal/templates/templates.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"bytes"
+	"fmt"
+	"strconv"
+	"text/template"
+
+	osbuilderprojectflottaiov1alpha1 "github.com/project-flotta/osbuild-operator/api/v1alpha1"
+)
+
+func Process(textTemplate string, expectedParameters []osbuilderprojectflottaiov1alpha1.Parameter, values []osbuilderprojectflottaiov1alpha1.ParameterValue) (string, error) {
```
ygalblum, line 12: As part of the controller for the OSBuildEnvConfig, I'm also using text templates in order to load configuration and K8S resource files (https://github.com/ygalblum/osbuild-operator/blob/osbuildenvconfig/internal/utils/template.go). So, to avoid collisions, let's discuss where to put the file (I used a general `utils` package) and choose meaningful function names.

jakub-dzon: Thanks, I was hoping for that kind of early review;) 
Yeah, it would be good to have all of that in one place.  All of our functions will form quite strong and cohesive core functionality of the operator and I think it deserves a dedicated package in contrast to `utils` which should rather be a collection of helpers. 
WDYT?

PR #28, opened 2022-05-30
Add IsoPackage command

eloycoto:
To be able to repackage a iso image, a new command was added to this
operator. This will run on a kubernetes Job, where all the the config
will be in there and pushed to to a new iso file.

The example output for the command is the following:

```
$ go run ./cmd/iso_package \
  -s http://localhost:8000/Fedora.iso \
  -ks test.cfg \
  -arch x86
INFO: 2022/05/30 17:53:22 Creating temp folder: /tmp/iso_package_tmp329664038
INFO: 2022/05/30 17:53:22 Writing iso image to '/tmp/iso_package_tmp329664038/original.iso'
INFO: 2022/05/30 17:53:24 Iso created correctly at /tmp/iso_package_tmp329664038/final.iso
```

Fixed: https://issues.redhat.com/browse/ECOPROJECT-767

Signed-off-by: Eloy Coto <eloy.coto@acalustra.com>

PR #37, opened 2022-06-03
Controller: Iso Packaging

eloycoto:
This is the controller job to create the ISO Package and upload it to
an AWS S3 endpoint.

This commits needs commits from PR #28, where it adds the iso_package
command to the platform.

Related to: https://issues.redhat.com/browse/ECOPROJECT-767
Related to: https://issues.redhat.com/browse/ECOPROJECT-831

Question regarding this PR: 
- [ ] Where the source image for the job is defined, configmap?
- [ ] Is the Kickstart file already part of any CM?
- [ ] AWS secrets need to be mounted from any secret, is that available at any
place?

The code that runs this is like this: 

```golang
package main

import (
	"fmt"
	"log"
	"time"

	"github.com/project-flotta/osbuild-operator/internal/iso_packaging"
	ctrl "sigs.k8s.io/controller-runtime"
	"sigs.k8s.io/controller-runtime/pkg/client"
)

var (
	defaultNamespace = "default"
)

func main() {
	config, err := ctrl.GetConfig()
	if err != nil {
		log.Fatalf("Failed: %v", err)
	}

	c, err := client.New(config, client.Options{})
	if err != nil {
		log.Fatalf("Failed: %v", err)
	}
	builder := iso_packaging.NewBuilderJob(c, "test", "default")
	
	_, err = builder.Start()
	
	status, err := builder.IsFinished()
	fmt.Println(status, err)

	time.Sleep(30 * time.Second)
	status, err = builder.IsFinished()
	fmt.Println(status, err)

	err = builder.Delete()
	fmt.Println(err)
}
```


ygalblum:
> Where the source image for the job is defined, configmap?
I think we should use the [ControllerManagerConfig](https://github.com/project-flotta/osbuild-operator/blob/main/config/manager/controller_manager_config.yaml) structure. I found this https://book.kubebuilder.io/component-config-tutorial/custom-type.html to explain how it can be done. The advantage of using it is that assuming we will use the operator's container image, we can use kustomize to update the value

> Is the Kickstart file already part of any CM?
The build process is handled by the controller of the OSBuild controller. The CRD of the OSBuild includes the Kickstart file: https://github.com/project-flotta/osbuild-operator/blob/main/api/v1alpha1/osbuild_types.go#L32

> AWS secrets need to be mounted from any secret, is that available at any place?
The S3 configuration information is provided as part of the the OSBuildEnvConfig CR: https://github.com/project-flotta/osbuild-operator/blob/972db79e056fbdce911108c79ebdfa78ce546763/api/v1alpha1/osbuildenvconfig_types.go#L101

eloycoto:
PR Ready for review: 

I had an offline chat with @ygalblum  and we agreed on the following:

- At the moment one runner test is enough, this code may change in a short period of time and do not want to waste time. https://issues.redhat.com/browse/ECOPROJECT-853
- There are some open questions if the Delete/IsFinished should be a specific task on OsBuild controller, or belongs here. 
- Maybe using the repository as k4e-operator.

internal/iso_packaging/iso_packaging.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+)
+
+const (
+	templateCommand string = "isopackage %s -ks %s --upload-target"
+	kickstart       string = "init.ks"
+)
+
+var (
+	jobTTLafterFinish int32 = 100   // Job will be deleted after finished with this
```
ygalblum, line 25: Sorry for nitpicking here. It should be a capital A: `jobTTLAfterFinish`

internal/iso_packaging/iso_packaging.go
```diff
@@ -0,0 +1,241 @@
...
+	if sourceISO == "" {
+		return fmt.Errorf("Cannot parse invalid iso image")
+	}
+	_, err := url.Parse(sourceISO)
+	if err != nil {
+		return err
+	}
+
+	if b.build.Spec.Kickstart == nil {
+		return fmt.Errorf("Kickstart is not defined, skipping")
```
ygalblum, line 64: I'm thinking out load here. Should this be an error or a noop?

internal/iso_packaging/iso_packaging.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	command := fmt.Sprintf(templateCommand, b.build.Status.ComposerIso, kickstart)
+
+	jobSpec := &batchv1.Job{
```
ygalblum, line 69: Consider moving this struct to a separate method to improve this one's readability.

internal/iso_packaging/iso_packaging.go
```diff
@@ -0,0 +1,241 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+
+	if b.buildConfig.Spec.S3Service.AWS != nil {
+
+		config := b.buildConfig.Spec.S3Service.AWS
```
ygalblum, line 124: Consider moving to a separate method to improve readability

internal/iso_packaging/iso_packaging.go
```diff
@@ -0,0 +1,241 @@
...
+			},
+		}
+
+		jobSpec.Spec.Template.Spec.Containers[0].Command = append(
+			jobSpec.Spec.Template.Spec.Containers[0].Command,
+			gets3Target(config.Bucket, b.build))
+	}
+
+	if b.buildConfig.Spec.S3Service.GenericS3 != nil {
+		config := b.buildConfig.Spec.S3Service.GenericS3
```
ygalblum, line 146: Consider moving to a separate method to improve readability

internal/iso_packaging/iso_packaging.go
```diff
@@ -0,0 +1,241 @@
...
+					},
+				})
+		}
+
+		jobSpec.Spec.Template.Spec.Containers[0].Command = append(
+			jobSpec.Spec.Template.Spec.Containers[0].Command,
+			gets3Target(config.Bucket, b.build))
+	}
+
+	err = b.client.Create(context.TODO(), jobSpec)
```
ygalblum, line 196: Maybe the `Start` method should get the context from the caller. WDYT?

internal/iso_packaging/iso_packaging.go
```diff
@@ -0,0 +1,241 @@
...
+			},
+		}
+		jobSpec.Spec.Template.Spec.Containers[0].Env = []v1.EnvVar{
+			{
+				Name:  "AWS_DEFAULT_REGION",
+				Value: config.Region,
+			},
+		}
+
+		jobSpec.Spec.Template.Spec.Containers[0].Command = append(
```
ygalblum, line 140: The `append` call for the `Command` is the same for both cases

internal/iso_packaging/iso_packaging.go
```diff
@@ -0,0 +1,241 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	if b.buildConfig.Spec.S3Service.AWS != nil {
```
ygalblum, line 122: Since we force an upload, these `if` statements should be `if` `else` with a failure at the end

internal/iso_packaging/iso_packaging.go
```diff
@@ -0,0 +1,241 @@
...
+		jobSpec.Spec.Template.Spec.Containers[0].Command = append(
+			jobSpec.Spec.Template.Spec.Containers[0].Command,
+			gets3Target(config.Bucket, b.build))
+	}
+
+	err = b.client.Create(context.TODO(), jobSpec)
+	if err != nil {
+		return fmt.Errorf("Cannot applied job: %v", err)
+	}
+	b.jobSpec = jobSpec
```
ygalblum, line 200: Do you really need to store the jobSpec? or just the name and namespace?

internal/iso_packaging/iso_packaging.go
eloycoto: :+1:

internal/iso_packaging/iso_packaging.go
eloycoto: error, if no kickstart shouldn't be here at all.

internal/iso_packaging/iso_packaging.go
eloycoto: No, if AWS is not in there does not mean that Generics3 can be in there, because is a pointer.

internal/iso_packaging/iso_packaging.go
eloycoto: :+1:

internal/iso_packaging/iso_packaging.go
eloycoto: Looks similar, but are different: 

Config is retrieved here: 
https://github.com/eloycoto/osbuild-operator/blob/55f89a026a4a4c7feb921daf342e6d30ed976b30/internal/iso_packaging/iso_packaging.go#L124

And here: 

https://github.com/eloycoto/osbuild-operator/blob/55f89a026a4a4c7feb921daf342e6d30ed976b30/internal/iso_packaging/iso_packaging.go#L146

internal/iso_packaging/iso_packaging.go
eloycoto: :+1:

internal/iso_packaging/iso_packaging.go
eloycoto: at this point I would like to store, I do not think that have heavy implications like memory at this stage. but happy to remove.

internal/iso_packaging/iso_packaging.go
ygalblum: OK, fine by me. But, maybe rephrase the message to show that it's an error. `Skipping` is usually used for noop

internal/iso_packaging/iso_packaging.go
ygalblum: You're right, my bad

internal/iso_packaging/iso_packaging.go
ygalblum: I meant `if AWS` ... `else if S3` ... `else error`. I think that if we don't return an error on the third case, the Job will fail instead

internal/iso_packaging/iso_packaging.go
ygalblum: You agreed but did not change the code :)

eloycoto: :facepalm:

bardielle: @eloycoto did you fix it? 
it seems that is still using context.TODO

eloycoto: yes, yes, I'll update today

eloycoto:
@ygalblum @bardielle this should be ready to merge it, sorry for the delay

config/manager/controller_manager_config.yaml
```diff
@@ -9,3 +9,5 @@ webhook:
 leaderElection:
   leaderElect: true
   resourceName: bfdcaedc.osbuilder.project-flotta.io
+isoPackage:
```
ygalblum, line 12: This change is not relevant anymore since this file it not really used

internal/conf/config.go
```diff
@@ -31,6 +31,9 @@ type OperatorConfig struct {
 
 	// CAIssuerName is the name of the cert-manager issuer in the operator's namespace used for the environment setup
 	CAIssuerName string `envconfig:"CA_ISSUER_NAME" required:"true"`
+
+	// BaseISOContainerImage is the container image to run the iso-package job
+	BaseISOContainerImage string `envconfig:"BASE_ISO_CONTAINER_IMAGE" required:"true"`
```
ygalblum, line 36: This change breaks the operator since you're not setting BASE_ISO_CONTAINER_IMAGE anywhere. It should be set either in the `controller_manager_config.properties` or directly in the PodSpec

ygalblum (changes requested):

config/manager/controller_manager_config.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: controller-runtime.sigs.k8s.io/v1alpha1
```
ygalblum, line 1: This file is not used anywhere so it was removed since you've opened the PR. Please remove it from the PR

main.go
```diff
@@ -82,6 +82,7 @@ func init() {
 }
 
 func main() {
+
```
ygalblum, line 85: Did you add this line intentionally?

internal/conf/config.go
```diff
@@ -58,6 +58,9 @@ type OperatorConfig struct {
 
 	// WorkerOSBuildVersion is the release tag of OSBuild for the osbuild RPM
 	WorkerOSBuildVersion string `envconfig:"WORKER_OSBUILD_VERSION" default:"v59"`
+
+	// BaseISOContainerImage is the container image to run the iso-package job
+	BaseISOContainerImage string `envconfig:"BASE_ISO_CONTAINER_IMAGE" required:"true" default:"controller:latest"`
```
ygalblum, line 63: The problem with this default value is that it is always incorrect.

You should set the environment variable to the name of the container image. See https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-container-fields-as-values-for-environment-variables for information

ygalblum (changes requested):

ygalblum (approved):

ygalblum (changes requested):
The file `controller_manager_config.yaml` should be removed

ygalblum (approved):

PR #38, opened 2022-06-05
Makefile: Add ginkgo test-fast

eloycoto:
This commits depends on PR #28, that adds ginkgo v2 support.

This is a small change on Makefile to run ginkgo directly, so with this
change, in future changes when Kind/e2e jobs are in place, we can use
gingko labels[0] to execute the right test[1].

[0] https://onsi.github.io/ginkgo/#spec-labels
[1] go test runner does not support ginkgo labels.


Makefile
```diff
@@ -306,3 +311,10 @@ catalog-build: opm ## Build a catalog image.
 .PHONY: catalog-push
 catalog-push: ## Push a catalog image.
 	$(MAKE) docker-push IMG=$(CATALOG_IMG)
+
+GINKGO = $(shell pwd)/bin/ginkgo
+.PHONY: ginkgo
+ginkgo: ## Download ginkgo locally if necessary.
+ifeq (, $(wildcard $(GINKGO)))
+	$(call go-install-tool,$(GINKGO),github.com/onsi/ginkgo/v2/ginkgo@v2.1.3)
```
bardielle, line 319: Please update also go.mod file to include ginkgo version 2

eloycoto: As I mentioned, the go.mod is part of the PR #28

bardielle: I think it's better to include it as part of that PR, I have a  PR with UT that includes changes in the Makefile, vendor, and go.mod file

ygalblum: V2 is included from #28 and removed here. @bardielle can we consider this conversion resolved?

ygalblum (approved):

PR #41, opened 2022-06-07
CRD: Using Kickstart from ConfigMap on OsBuild

eloycoto:
Because this is an snapshot of what template is doing, the kickstart is
on a Configmap because can be used to open the job and merged into a
container to be used by cmd/iso_package command.

The main reason why template is not using CM is because is what the user
provided, and does not make sense at all.


config/crd/bases/osbuilder.project-flotta.io_osbuilds.yaml
```diff
@@ -131,9 +131,17 @@ spec:
...
               kickstart:
-                description: Kickstart may store content of a kickstart file to be
-                  used in the target image
-                type: string
+                description: Kickstart is a reference to a configmap that may store
+                  content of a kickstart file to be used in the target image
+                properties:
+                  name:
+                    description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                      TODO: Add other useful fields. apiVersion, kind, uid?'
```
bardielle, line 139: this is an auto-generated file by the command `make manifests`
Please edit the comment on top of the kickstart field

bardielle (approved):

PR #45, opened 2022-06-08
Implement admission web-hooks for OSBuildEnvConfig

ygalblum:
For now OSBuildEnvConfig is an immutable singleton CR. As such, only one can exist at any point and it cannot be updated.

Note, the first commit is auto-generated code.


api/v1alpha1/osbuildenvconfig_webhook.go
```diff
@@ -54,7 +77,17 @@ var _ webhook.Validator = &OSBuildEnvConfig{}
 func (r *OSBuildEnvConfig) ValidateCreate() error {
 	osbuildenvconfiglog.Info("validate create", "name", r.Name)
 
-	// TODO(user): fill in your validation logic upon object creation.
+	ctx := context.Background()
+	osBuildEnvConfigList := OSBuildEnvConfigList{}
+	err := kClient.List(ctx, &osBuildEnvConfigList)
```
bardielle, line 82: Do you think that wrapping it with a new repository can help later? (maybe for the controller)

ygalblum: I do and I have added it for the controller. However, I couldn't use it here as it created a cyclic import. I could create an intermediate package but I think it's an overkill

bardielle: not sure what is better - both ways are sound good

bardielle (approved):

api/v1alpha1/osbuildenvconfig_webhook.go
ygalblum: I think adding an intermediate package just for the sake of importing for such a small piece of code is over engineering. I'll leave it as is

PR #47, opened 2022-06-08
OSBuildEnvConfig - Add Name to WorkerConfig and require at least one

ygalblum:
Add Name to the WorkerConfig struct
Define WorkersConfig type and enforce at least one item as the API level
Enforce uniqueness of the worker name in the webhook
Add unittest


api/v1alpha1/osbuildenvconfig_webhook.go
```diff
@@ -91,6 +108,17 @@ func (r *OSBuildEnvConfig) ValidateCreate() error {
 	return nil
 }
 
+func validateUniqueWorkerNames(workers []WorkerConfig) error {
+	workerNames := make(map[string]bool)
```
jakub-dzon, line 112: Not that is a big deal but `map[string]struct{}` has better memory utilization (`struct{}{}` doesn't take any memory),

ygalblum: Sure. What do I set when adding a key to the map? `workerNames[worker.Name] = ???`

ygalblum: Done

bardielle (approved):

jakub-dzon (approved):

PR #27, opened 2022-05-30
Add webhook trigger

bardielle:
This PR adds a rest API for `wehbhook triggers`, allowing you to trigger a new `OSBuild` CR by sending a request to the OCP endpoint. 
The new rest API is a different pod that runs separately from the operator. 

For all webhooks, you must define a secret with a key named `WebHookSecretKey` and the value being the value to be supplied when invoking the webhook.
The request should includes the secret value in the request's header and looks like:` curl -v -H "secret: secretvalue" -X POST http://<SERVICE_IP>:8080/api/osbuild/v1/namespaces/<osbuildconfig_namespace>/osbuildconfig/<osbuildconfig_name>/webhooks`

PR #48, opened 2022-06-08
Use OSBuildConfig template

jakub-dzon:
This PR introduces basic use of `OSBuildConfigTemplate` objects - changes to the template do not cause related `OSBuildConfig` reconciliation - yet.  That functionality will be introduced in subsequent PRs. 

For now new `OSBuild` will be created only when `OSBuildConfig` changes to limit the size of the change set.

The functionality should be covered with tests, preferably in (unmerged yet) [this test suite](https://github.com/project-flotta/osbuild-operator/pull/27/files#diff-ca193fb8e09bfc9b5de1cb790ac9acbb2b041beaf3064bdf5714fda26bb816fb) or tests for [`OSBuildCreator`](https://github.com/project-flotta/osbuild-operator/blob/013b38ab3ae0777ccc69cab004979d38ed45c692/internal/manifests/osbuild.go) - probably in a follow-up PR or here, after #27 is merged.

The changes have been tested manually with [sample flotta manifests](https://github.com/project-flotta/osbuild-operator/tree/15c51a78c5b8211314fd151c04b065cc55c9b89d/config/samples/flotta).


controllers/osbuildconfig_controller.go
```diff
@@ -125,11 +147,134 @@ func (r *OSBuildConfigReconciler) createNewOSBuildCR(ctx context.Context, osBuil
...
+		}
+
+		mergeCustomization(osConfigTemplate, osBuildConfigSpecDetails, buildConfigCustomizations)
+
+		kickstartConfigMap, err = r.createKickstartConfigMap(ctx, osBuildConfig, osConfigTemplate, osBuildName, osBuild.Namespace)
+		if err != nil {
+			return nil, nil, err
+		}
+		if kickstartConfigMap != nil {
+			osBuild.Spec.Kickstart = &osbuilderprojectflottaiov1alpha1.NameRef{Name: osBuildName}
```
bardielle, line 181: Maybe the name needs to be kickstartConfigMap's name? (here in your case it's "osBuildName" )

bardielle (approved):
This code looks good, I think it important to add a UT (a specially to the template processing functions) 
Can you open a ticket for that? (if it wasn't done yet)

jakub-dzon:
> This code looks good, I think it important to add a UT (a specially to the template processing functions) Can you open a ticket for that? (if it wasn't done yet)

Thanks for having a look. 
Yes - I think the best place for the test would be [here](https://github.com/project-flotta/osbuild-operator/blob/e873d09a85983ee8ec516ef3faac6cd50ba45d1b/internal/osbuildconfig/osbuildconfig_handler_test.go) or in [here](https://github.com/project-flotta/osbuild-operator/blob/e873d09a85983ee8ec516ef3faac6cd50ba45d1b/internal/manifests) - after your PR is merged.

I will create an issue for it so it wouldn't be forgotten.

controllers/osbuildconfig_controller.go
```diff
@@ -125,11 +147,134 @@ func (r *OSBuildConfigReconciler) createNewOSBuildCR(ctx context.Context, osBuil
...
+
+		err = r.Client.Create(ctx, cm)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return cm, nil
+}
+
+func mergeCustomization(osConfigTemplate *osbuilderprojectflottaiov1alpha1.OSBuildConfigTemplate, osBuildConfigSpecDetails *osbuilderprojectflottaiov1alpha1.BuildDetails, buildConfigCustomizations *osbuilderprojectflottaiov1alpha1.Customizations) {
```
ygalblum, line 231: Not sure I understand what this method will produce. Maybe we weren't clear about the expected output either.
For example, we would want to allow the user to add packages at the `OSBuildConfig` level to the ones defines at the `OSBuildConfigTemplate` level. E.g. if the `OSBuildConfigTemplate` sets packages `a` and `b`, the `OSBuildConfig` should be able to add package `c`. So, there should be a smart merge here (i.e. merge and unique).
- In `Packages` it is simple as it is a list of strings.
-  In Users we need to check the user's name and if exists on both, the `OSBuildConfig` takes precedence. 
- Services is the most complicated one as we need to make sure the `enabled` and `disabled` lists do not contradict each other. But the rule stays the same the `OSBuildConfig` takes precedence

jakub-dzon: I understood from our previous chat, that `OSBuildConfig` takes precedence and so it does here, but at the level of whole sets of customizations - packages, services and users. For example - if packages are configured in both template and build config, only packages from build config would be used. If that's not the desired behavior, I will rework it.

controllers/osbuildconfig_controller.go
```diff
@@ -125,11 +147,134 @@ func (r *OSBuildConfigReconciler) createNewOSBuildCR(ctx context.Context, osBuil
...
+			return nil, nil, err
+		}
+
+		mergeCustomization(osConfigTemplate, osBuildConfigSpecDetails, buildConfigCustomizations)
+
+		kickstartConfigMap, err = r.createKickstartConfigMap(ctx, osBuildConfig, osConfigTemplate, osBuildName, osBuild.Namespace)
+		if err != nil {
+			return nil, nil, err
+		}
+		if kickstartConfigMap != nil {
```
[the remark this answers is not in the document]
jakub-dzon: Good catch! Thanks!

controllers/osbuildconfig_controller.go
jakub-dzon: Done.

internal/templates/templates.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"bytes"
+	"fmt"
+	"strconv"
+	"text/template"
+
+	osbuilderprojectflottaiov1alpha1 "github.com/project-flotta/osbuild-operator/api/v1alpha1"
+)
+
+func Process(textTemplate string, expectedParameters []osbuilderprojectflottaiov1alpha1.Parameter, values []osbuilderprojectflottaiov1alpha1.ParameterValue) (string, error) {
```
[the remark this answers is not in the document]
jakub-dzon: Done.

jakub-dzon:
@ygalblum - please have another look, I've addressed your comments.

controllers/suite_test.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package controllers
+package controllers_test
```
ygalblum, line 17: What's the best practice here - put the tests in the same package or *_test? I don't mind it, just want to be consistent

internal/customizations/customizations.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"github.com/project-flotta/osbuild-operator/api/v1alpha1"
+)
+
+var void struct{}
+
+func MergeCustomizations(templateCustomizations, configCustomizations *v1alpha1.Customizations) *v1alpha1.Customizations {
+	var customizations *v1alpha1.Customizations
+	if configCustomizations != nil {
```
ygalblum, line 11: I think this block (including the `if`) should be the `else` block of the next `if` because you override `customizations` anyhow

controllers/suite_test.go
jakub-dzon: `_test` so we are not tempted to test non-public APIs and write sensible public ones.

controllers/osbuildconfig_controller.go
```diff
@@ -19,35 +19,44 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"github.com/project-flotta/osbuild-operator/internal/customizations"
```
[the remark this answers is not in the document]
jakub-dzon: Done.

internal/customizations/customizations.go
jakub-dzon: Yeah, that will work too.

controllers/osbuildconfig_controller.go
```diff
@@ -19,35 +19,44 @@ package controllers
...
+	"time"
 
 	"github.com/go-logr/logr"
+	v1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
jakub-dzon: Done.

ygalblum (approved):

PR #51, opened 2022-06-09
Fixing OSBuildConfigTemplate controller problems

jakub-dzon:
This PR fixes problems with `OSBuildConfigTemplate` caused by wrong scaffolding.

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -5,12 +5,6 @@ generatorOptions:
   disableNameSuffixHash: true
 
 configMapGenerator:
-- files:
+- name: manager-config
+  files:
   - controller_manager_config.yaml
-  name: manager-config
-apiVersion: kustomize.config.k8s.io/v1beta1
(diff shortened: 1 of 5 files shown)


config/manager/kustomization.yaml
```diff
@@ -5,6 +5,12 @@ generatorOptions:
...
-- name: manager-config
-  files:
+- files:
   - controller_manager_config.yaml
+  name: manager-config
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: quay.io/jdzon/osbuild-operator
```
eloycoto, line 15: Shouldn't we use the project-flotta quay?

eloycoto (approved):

config/manager/kustomization.yaml
jakub-dzon: Fixed

pkliczewski (approved):

PR #53, opened 2022-06-10
Fix various problems with OSBuildConfig reconciliation

jakub-dzon:
This PR fixes various problems with `OSBuildConfig` reconciliation:
 - Panic when the CR is reconciled for the first time and  `osBuildConfig.Status.LastVersion` is `nil`;
 - Generated `OSBuild` name contained illegal characters;
 - Generated `OSBuild` didn't have namespace specified;
 - One `OSBuildConfig` was reconciled endlessly because we changed its status, which triggered another reconciliation. As a result new `OSBuild` objects were repeatedly created.
 - Logger panicked, because it received odd number of parameters (it should message followed by any number of <key, value> pairs).

### DIFF
--- a/controllers/osbuildconfig_controller.go
+++ b/controllers/osbuildconfig_controller.go
@@ -18,10 +18,10 @@ package controllers
 
 import (
 	"context"
+	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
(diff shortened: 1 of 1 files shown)


pkliczewski (approved):
Would be great to add some tests in the future.

PR #56, opened 2022-06-14
Operator Configuration - support environment variables based configuration

ygalblum:
# Operator Configuration - support environment variables based configuration

## Internal conf package
Load configuration from env using envconfig
Implement configuration validate and get

## main.go
Load all the configuration using the new package
Remove previously defined flags
Get the configuration values from the conf pacakge
Create the logger based on the configuration

## Manager Deployment
Add a properties file to set the environment variables
Create a configMap from the properties file
Get environment variables from the configMap
Pass the operator namespace using the environment variable resolved from the deployment spec
Do not pass any arguments as the values will be taken from the env

## Vendor
Add github.com/kelseyhightower/envconfig as a dependency

## Misc
Ignore vscode directory


jakub-dzon:
> > Why is there a need to load configuration from any other source than env variables? It looks way more complicated than simple `envconfig` usage.
> 
> The operator SDK already creates the configMap and I think that a configuration file wrapped in a ConfigMap is cleaner than environment variables. Plus this implementation does not break other configuration options (env or flags)

That config map can be used as env source for the container - why not use it in a simple way, like in the Flotta Operator? I don't think there is a need for additional configuration abstraction (indirection) in the shape of a file wrapped in a ConfigMap.

ygalblum:
> > > Why is there a need to load configuration from any other source than env variables? It looks way more complicated than simple `envconfig` usage.
> > 
> > 
> > The operator SDK already creates the configMap and I think that a configuration file wrapped in a ConfigMap is cleaner than environment variables. Plus this implementation does not break other configuration options (env or flags)
> 
> That config map can be used as env source for the container - why not use it in a simple way, like in the Flotta Operator? I don't think there is a need for additional configuration abstraction (indirection) in the shape of a file wrapped in a ConfigMap.

But the mechanism for the ConfigMap is already in place, I'm not inventing anything. Plus, that configMap is a single entry one designed to be a file, not parsed into envs

jakub-dzon:
Please have a look at Flotta operator what I have in mind instead:
 - [config map generator](https://github.com/project-flotta/flotta-operator/blob/fe100a1eaa74c82ab12bc3adf1ade6441bcef7f7/config/manager/kustomization.yaml#L7) creating default CM with env variables for the operator
 - [use of the map](https://github.com/project-flotta/flotta-operator/blob/fe100a1eaa74c82ab12bc3adf1ade6441bcef7f7/config/manager/manager.yaml#L34)
 - [config model](https://github.com/project-flotta/flotta-operator/blob/fe100a1eaa74c82ab12bc3adf1ade6441bcef7f7/cmd/manager/main.go#L77)
 - [config model loading](https://github.com/project-flotta/flotta-operator/blob/fe100a1eaa74c82ab12bc3adf1ade6441bcef7f7/cmd/manager/main.go#L127) - one line instead of byzantine (in comparison) use of Viper.

ygalblum:
> Please have a look at Flotta operator what I have in mind instead:
> 
> * [config map generator](https://github.com/project-flotta/flotta-operator/blob/fe100a1eaa74c82ab12bc3adf1ade6441bcef7f7/config/manager/kustomization.yaml#L7) creating default CM with env variables for the operator
> * [use of the map](https://github.com/project-flotta/flotta-operator/blob/fe100a1eaa74c82ab12bc3adf1ade6441bcef7f7/config/manager/manager.yaml#L34)
> * [config model](https://github.com/project-flotta/flotta-operator/blob/fe100a1eaa74c82ab12bc3adf1ade6441bcef7f7/cmd/manager/main.go#L77)
> * [config model loading](https://github.com/project-flotta/flotta-operator/blob/fe100a1eaa74c82ab12bc3adf1ade6441bcef7f7/cmd/manager/main.go#L127) - one line instead of byzantine (in comparison) use of Viper.

I see your point. One thing though, I think I would still use a separate package for the configuration so that it is accessible to other parts of the code (and not through the main package). WDYT?
BTW in flotta-operator, do you use the [controller_manager_config.yaml](https://github.com/project-flotta/flotta-operator/blob/fe100a1eaa74c82ab12bc3adf1ade6441bcef7f7/config/manager/controller_manager_config.yaml) file?

jakub-dzon:
> I see your point. One thing though, I think I would still use a separate package for the configuration so that it is accessible to other parts of the code (and not through the main package). WDYT? 

Strong yes for that part - I've done [something similar](https://github.com/project-flotta/flotta-operator/blob/main/internal/edgeapi/config.go) for the edgeapi.

> BTW in flotta-operator, do you use the [controller_manager_config.yaml](https://github.com/project-flotta/flotta-operator/blob/fe100a1eaa74c82ab12bc3adf1ade6441bcef7f7/config/manager/controller_manager_config.yaml) file?

That's a very good question - it seems to be unused and probably should be removed.

ygalblum:
@jakub-dzon Done

internal/conf/config.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+var globalConf *OperatorConfig
+
+// Get the global operator configuration object.
+func Get() *OperatorConfig {
+	return globalConf
+}
+
+func (oc *OperatorConfig) Validate() error {
```
jakub-dzon, line 42: WDYT about having logic from `Validate` to be integral part of `Load` (not inline, but a call to private version of `Validate`) so there would be no need for another call and possibility of forgetting about it during some refactor? 

The result of `Load` failing is exactly the same as for `Validate` - app exit.

main.go
```diff
@@ -58,29 +53,37 @@ func init() {
...
+	setupLog.Info("Started with configuration", "configuration", conf.Get())
+
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme,
-		MetricsBindAddress:     metricsAddr,
-		Port:                   9443,
-		HealthProbeBindAddress: probeAddr,
-		LeaderElection:         enableLeaderElection,
-		LeaderElectionID:       "bfdcaedc.osbuilder.project-flotta.io",
+		MetricsBindAddress:     conf.Get().MetricsAddr,
```
jakub-dzon, line 82: Please create a `configuration:=conf.Get()` and use it throughout the method. I think it would be more readable.

internal/conf/config.go
```diff
@@ -0,0 +1,59 @@
...
+	WorkingNamespace string `envconfig:"WORKING_NAMESPACE" default:""`
+
+	// Verbosity of the logger.
+	LogLevel string `envconfig:"LOG_LEVEL" default:"info"`
+}
+
+var globalConf *OperatorConfig
+
+// Get the global operator configuration object.
+func Get() *OperatorConfig {
```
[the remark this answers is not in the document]
ygalblum: You're right, done

internal/conf/config.go
ygalblum: Sure, makes sense - Done

main.go
ygalblum: Once I externalized the `GlobalConf` this becomes less relevant, right?

jakub-dzon: Yeah, it's good now.

PR #60, opened 2022-06-15
Use cmctl to wait for the cert-manager operator to become ready

ygalblum:
`make deploy` usually fails because the operator's manifests are deployed before the cert-manager operator is ready. The cert-manager [documentation](https://cert-manager.io/docs/installation/verify/#check-cert-manager-api) suggests verifying the installation using `cmctl`. So this change uses `cmctl` instead of just waiting for the Pods to become ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -217,9 +217,9 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 	$(KUSTOMIZE) build config/crd | $(CLUSTER_RUNTIME) delete --ignore-not-found=$(ignore-not-found) -f -
 
 .PHONY: install-cert-manager
-install-cert-manager:
+install-cert-manager: cmctl
 	$(CLUSTER_RUNTIME) apply -f https://github.com/cert-manager/cert-manager/releases/download/${CERT_MANAGER_VERSION}/cert-manager.yaml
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -246,6 +246,16 @@ ENVTEST = $(shell pwd)/bin/setup-envtest
 envtest: ## Download envtest-setup locally if necessary.
 	$(call go-install-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest@latest)
 
+CMCTL = $(shell pwd)/bin/cmctl
+OS = $(shell go env GOOS)
+ARCH = $(shell go env GOARCH)
+.PHONY: cmctl
+cmctl: ## Download cmctl locally if necessary.
+ifeq ("$(wildcard $(CMCTL))","")
```
bardielle, line 254: I'm preferring to use go install: 
`go install github.com/oleewere/cmctl@v0.5.0`

ygalblum: I'm following the installation steps as documented in the tools documentation.
Furthermore, while it's name is the same, the repo you provided is a totally different tool that has nothing to do with the certificate manager

bardielle: Yes, I saw those instructions but I tried to find a go install cmd for cmctl.
So I guess It's fine by me

bardielle (approved):

PR #63, opened 2022-06-17
Reconcile config on template change

jakub-dzon:
This PR finalizes `OSBuildConfigTemplate` support  by adding following changes:
 - new flag `OSBuildConfig.spec.triggers.templateConfigChange` for specifying whether associated template change should trigger build creation;
 - updating `OSBuildConfig` with latest version of associated `OSBuildConfigTemplate` when template changes.


internal/predicates/osbuildconfig.go
```diff
@@ -29,5 +29,9 @@ func (OSBuildConfigChangedPredicate) Update(e event.UpdateEvent) bool {
 	if newConfig.Status.LastTemplateResourceVersion != nil && newConfig.Status.CurrentTemplateResourceVersion != nil {
 		templateChanged = *newConfig.Status.LastTemplateResourceVersion != *newConfig.Status.CurrentTemplateResourceVersion
 	}
+
+	templateTriggerEnabled := newConfig.Spec.Triggers.TemplateConfigChange == nil || *newConfig.Spec.Triggers.TemplateConfigChange
```
ygalblum, line 33: If we're checking if template change should trigger a new build, we should also check if a new config should trigger a new build (for the generationChanged flag)

internal/predicates/osbuildconfig_test.go
```diff
@@ -49,6 +52,23 @@ var _ = Describe("OSBuildConfig reconciliation predicate", func() {
 				},
 			},
 		),
+		Entry("when template version changed and template trigger is explicitly enabled",
```
ygalblum, line 55: Should we also check when the template trigger is not set (default behavior)?

jakub-dzon: It's covered here: https://github.com/project-flotta/osbuild-operator/pull/63/files#diff-4d944c176d5a7c6d7e5a0e1c0bd2ba6bd49d8e0479a6d8f225840691d81b8ffaR35

internal/predicates/osbuildconfig.go
jakub-dzon: :+1:

jakub-dzon: Done

ygalblum (approved):

PR #69, opened 2022-06-20
Use cert-manager CRDs to manage osbuild environment certificates

ygalblum:
Certificates
------------
Generate a self-signed CA issuer
Generate a certificate for the operator
Until rotation is handled, set duration to 10 years

Operator
--------
Attach the operator's certificate to its pod
Pass the name of the CA issuer to the operator


internal/conf/config.go
```diff
@@ -40,6 +43,12 @@ func (oc *OperatorConfig) validate() error {
 	if oc.WorkingNamespace == "" {
 		return fmt.Errorf("WorkingNamespace value [%s] invalid", oc.WorkingNamespace)
 	}
+
+	// Ensure that CAIssuerName is set.
+	if oc.CAIssuerName == "" {
+		return fmt.Errorf("CAIssuerName value [%s] invalid", oc.CAIssuerName)
```
jakub-dzon, line 49: There is no need to manually validate that; you can replace `default` with `required:"true"` in the struct tag for `CAIssuerName`.
And when we're at it, the same can be done with `WorkingNamespace`. That will remove the need for `validate`.

ygalblum: I didn't know about it, thanks! Done

jakub-dzon (approved):

PR #73, opened 2022-06-21
OSBuildEnvConfig controller - first phase

ygalblum:
The first phase focuses on starting the OSBuild-Composer Service:
1. Generate a Certificate
2. Create the ConfigMap for the service's configuration
3. Create the ConfigMap for the proxy's configuration
4. Create the Deployment for the OSBuild-Composer
5. Create the Services (Composer and Worker APIs)


internal/repository/osbuildenvconfig/osbuildenvconfig.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/project-flotta/osbuild-operator/api/v1alpha1"
+)
+
+//go:generate mockgen -package=osbuildenvconfig -destination=mock_osbuildenvconfig.go . Repository
+type Repository interface {
+	Read(ctx context.Context, name string) (*v1alpha1.OSBuildEnvConfig, error)
+	Update(ctx context.Context, osBuildEnvConfig *v1alpha1.OSBuildEnvConfig) error
```
jakub-dzon, line 16: Did you consider using `Patch`? It should be "safer" than `Update`, because `Update` will fail when the edited resource isn't fresh (was modified concurrently) and patch will try to merge the resources. 
I am not saying `Patch` is better than `Update`, but they have a bit different applications and behavior.

controllers/osbuildenvconfig_controller.go
```diff
@@ -18,24 +18,109 @@ package controllers
...
+	ComposerCertsSecretName          string
+	ComposerCertsSecretPublicCertKey string
+	ComposerCertsSecretPrivateKeyKey string
+	ComposerCertsSecretCACertKey     string
+	ProxyConfigMapName               string
+}
+
 // OSBuildEnvConfigReconciler reconciles a OSBuildEnvConfig object
 type OSBuildEnvConfigReconciler struct {
 	client.Client
```
jakub-dzon, line 106: Is the `Client` needed? All data management seems to be delegated to repositories.

controllers/osbuildenvconfig_controller.go
```diff
@@ -47,16 +132,417 @@ type OSBuildEnvConfigReconciler struct {
...
+	if instance.Spec.Composer != nil && instance.Spec.Composer.PSQL != nil {
+		composerDeploymentParams.PGSQLSecretName = instance.Spec.Composer.PSQL.ConnectionSecretReference.Name
+		if instance.Spec.Composer.PSQL.SSLMode != nil {
+			composerDeploymentParams.PgSSLMode = string(*instance.Spec.Composer.PSQL.SSLMode)
+		}
+	} else {
+		return ctrl.Result{Requeue: true}, fmt.Errorf("creating a PSQL service is not yet implemented")
+	}
+
+	_, changed, err = r.getOrCreateComposerConfigMap(ctx, instance)
```
jakub-dzon, line 230: Each step of the `Update` will be executed in separate reconciliation - did you consider creating as many resources as possible during one reconciliation? That would reduce time needed to create an environment.
The code now looks clean and easy to understand - would that be severely impacted by that kind of change?

controllers/osbuildenvconfig_controller.go
ygalblum: Good question. The `Client` is auto generated by the SDK for all controllers. So, I think we should consider removing it for all of them. But, I think that should be in a separate PR.
Should I remove it here?

jakub-dzon: Please remove this one here and let's do the rest separately.

controllers/osbuildenvconfig_controller.go
jakub-dzon: I don't have strong opinion here. I think this pattern would work good (better than all in one go) if accompanied by status conditions reflecting progress. Do you have plans for adding them?

ygalblum: Yes, I am planning on adding progress events and update the conditions in the status

internal/repository/osbuildenvconfig/osbuildenvconfig.go
ygalblum: Done

jakub-dzon (approved):

controllers/osbuildenvconfig_controller.go
```diff
@@ -47,16 +130,419 @@ type OSBuildEnvConfigReconciler struct {
...
+
+	changed, err := r.addFinalizer(ctx, instance)
+	if err != nil {
+		return ctrl.Result{Requeue: true}, err
+	} else if changed {
+		reqLogger.Info("Added finalizer")
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	_, changed, err = r.getOrCreateCertificate(
```
bardielle, line 180: it seems that this is the only place for using that method, so maybe need to change the name and the return params (not include the certs)

ygalblum: It's true for other methods as well as this is a pattern of getOrCreate. As a pattern I think I like keeping it constant and decide whether I want to use the returned value or not at the caller side

controllers/osbuildenvconfig_controller.go
```diff
@@ -47,16 +130,419 @@ type OSBuildEnvConfigReconciler struct {
...
+		return ctrl.Result{Requeue: true}, err
+	} else if changed {
+		reqLogger.Info("Created composer certificate")
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	composerDeploymentParams := composerDeploymentParameters{
+		ComposerDeploymentNamespace:      conf.GlobalConf.WorkingNamespace,
+		ComposerDeploymentName:           composerDeploymentName,
+		ComposerImageName:                "quay.io/app-sre/composer",
```
bardielle, line 200: I think it can be useful to export those values to consts

internal/templates/templates.go
```diff
@@ -54,3 +60,23 @@ func validateParameter(value, pType string) bool {
 		return true
 	}
 }
+
+func LoadFromTemplate(templateFilename string, params interface{}) (*bytes.Buffer, error) {
```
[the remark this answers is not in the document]
ygalblum: `LoadFromFile`? (since it's already in package `template`)? Or `LoadFromTemplateFile`?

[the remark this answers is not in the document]
bardielle: `LoadFromTemplateFile` sounds good

controllers/osbuildenvconfig_controller.go
ygalblum: Done

ygalblum:
@bardielle I also make the change we discussed regarding the getOrCreate... which are not ensure...Exists.

bardielle (approved):

PR #91, opened 2022-07-03
OSBuild controller create an edge-container job - first phase 

bardielle:
This PR focuses on creating a new edge-container composer job for each osbuild, and managing it.

Controller
-------
Starting a new composer job for creating a new edge-container image:
- send `postComposer` request with the OSBuild details
- Parse the response
- Update the OSBuild status with the composer job status
- For every 2 minutes update OSBuild status with composer job status 

Also, generate a composer client

Types
-------

- Adding the field `ContainerComposeId` and `IsoComposeId` to OSBuildStatus type 
- Updating the `OSBuildConditionType` values 
- Reordering the structs of `User` and `OSTreeConfig` of `OSBuildConfig` type to correspond the Composer structs

main.go
-------

- Add a compose client creation (mTLS)


controllers/osbuild_controller.go
```diff
@@ -18,19 +20,51 @@ package controllers
 
 import (
 	"context"
+	"fmt"
+	"github.com/go-logr/logr"
```
ygalblum, line 24: move github import to the middle section

controllers/osbuild_controller.go
```diff
@@ -18,19 +20,51 @@ package controllers
...
 
-	osbuilderprojectflottaiov1alpha1 "github.com/project-flotta/osbuild-operator/api/v1alpha1"
+	osbuildv1alpha1 "github.com/project-flotta/osbuild-operator/api/v1alpha1"
+	"github.com/project-flotta/osbuild-operator/internal/composer"
+	repositoryosbuild "github.com/project-flotta/osbuild-operator/internal/repository/osbuild"
+)
+
+const (
+	// Conditions Messages
+	FailedToSendPostRequestMsg      = "Failed to post a new composer build request"
```
ygalblum, line 41: Should all these consts be exposed?

controllers/osbuild_controller.go
```diff
@@ -18,19 +20,51 @@ package controllers
...
+	FailedIsoBuild        = "failedIsoBuild"
+	StartedIsoBuild       = "startedIsoBuild"
+
+	// Image types
+	EdgeContainerImgType = "edge-container"
+	EdgeInstallerImgType = "edge-installer"
+
+	EmptyComposeID = ""
+
+	RequeueForTwoMinutes = time.Minute * 2
```
ygalblum, line 60: Not sure you want to name the field bases on its value. What happens if we decide to change it to three minutes?

controllers/osbuild_controller.go
```diff
@@ -47,16 +81,220 @@ type OSBuildReconciler struct {
...
 func (r *OSBuildReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx).WithValues("osbuild", req.Name)
+
+	osBuild, err := r.OSBuildRepository.Read(ctx, req.Name, req.Namespace)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{Requeue: true, RequeueAfter: RequeueForTwoMinutes}, err
```
ygalblum, line 91: Why wait for two minutes in case of errors?

controllers/osbuild_controller.go
bardielle: In the next phase the controller will send a postRequest with edge-installer
I guess I'll use the same messages

controllers/osbuild_controller.go
bardielle: On  a different PR I got the opposite approach - that someone preferred to mentioned the specific time

controllers/osbuild_controller.go
```diff
@@ -47,16 +81,220 @@ type OSBuildReconciler struct {
...
+			return ctrl.Result{Requeue: true, RequeueAfter: RequeueForTwoMinutes}, nil
+		}
+	} else if composeStatus == composer.ComposeStatusValueSuccess && osBuild.Spec.Details.TargetImage.TargetImageType == EdgeInstallerImgType {
+		// TODO postComposeEdgeInstaller
+	}
 
 	return ctrl.Result{}, nil
 }
 
+func (r *OSBuildReconciler) UpdateContainerComposeId(ctx context.Context, logger logr.Logger, osBuild *osbuildv1alpha1.OSBuild) (composer.ComposeStatusValue, error) {
```
ygalblum, line 150: I think updateContainerComposeStatus is a better name since the method updates the status, not the id

ygalblum, line 150: Should all these methods be exposed?

controllers/osbuild_controller.go
```diff
@@ -47,16 +81,220 @@ type OSBuildReconciler struct {
...
+	if err != nil {
+		logger.Error(err, "failed to get compose ID status")
+		return "", err
+	}
+
+	r.updateOSBuildConditionStatus(ctx, logger, osBuild, composeStatus, ContainerBuildDone, FailedContainerBuild, StartedContainerBuild)
+	return composeStatus, nil
+}
+
+func (r *OSBuildReconciler) UpdateIsoComposeId(ctx context.Context, logger logr.Logger, osBuild *osbuildv1alpha1.OSBuild) (composer.ComposeStatusValue, error) {
```
ygalblum, line 161: Same about the name and exposure

controllers/osbuild_controller.go
bardielle: for retring

controllers/osbuild_controller.go
```diff
@@ -47,16 +81,220 @@ type OSBuildReconciler struct {
...
+	}
+
+	var composeStatus composer.ComposeStatusValue
+	containerBuildStatus := osbuildv1alpha1.OSBuildConditionType("")
+	if osBuild.Status.Conditions != nil {
+		conditionLen := len(osBuild.Status.Conditions)
+		containerBuildStatus = osBuild.Status.Conditions[conditionLen-1].Type
+	}
+
+	if osBuild.Status.ContainerComposeId != EmptyComposeID && containerBuildStatus != ContainerBuildDone {
```
ygalblum, line 107: I think there's a problem in the `if...else` statement.
First run, there is no Compose ID, so you go into the `else`
Second run, there is a Compose ID, but the Status is not Done so you go into the `if`.
Once the status is Done, you'll go into the `else` and reschedule another compose.
First check if ContainerComposeID is empty, if yes, schedule a build and return. If it's not empty, you can continue to check it's status. If it's not Done, you check and requeue until it is. Once it's done you'll continue to do the same for the ISO build

controllers/osbuild_controller.go
ygalblum: But, it will still be in the same controller, no?

controllers/osbuild_controller.go
```diff
@@ -47,16 +81,220 @@ type OSBuildReconciler struct {
...
+		r.updateOSBuildStatus(ctx, logger, osBuild, FailedToSendPostRequestMsg, FailedContainerBuild, EmptyComposeID, EmptyComposeID)
+		return err
+	}
+
+	composerResponse, err := composer.ParsePostComposeResponse(response)
+	if err != nil {
+		logger.Error(err, "failed parsing the response of postCompose")
+		return err
+	}
+	if composerResponse.StatusCode() != http.StatusCreated {
```
ygalblum, line 213: If the composer returned an error, this means that the Container build failed. So, the resource Status needs to be updated as well.

controllers/osbuild_controller.go
```diff
@@ -47,16 +81,220 @@ type OSBuildReconciler struct {
...
+func (r *OSBuildReconciler) createCustomizations(osbuildCustomizations osbuildv1alpha1.Customizations) composer.Customizations {
+	var users []composer.User
+	for _, cstmzUser := range osbuildCustomizations.Users {
+		user := (*composer.User)(cstmzUser.DeepCopy())
+		users = append(users, *user)
+	}
+
+	composerCustomizations := composer.Customizations{
+		Packages: &osbuildCustomizations.Packages,
+		Users:    &users,
```
ygalblum, line 290: The composer API was updated to support the service enable/disable list. You can add it either now, or in a separate PR

controllers/osbuildenvconfig_controller.go
```diff
@@ -90,6 +90,8 @@ const (
 	envoyProxyCertsDir = "/etc/certs"
 )
 
+var GlobalComposerComposerAPIServiceName = composerComposerAPIServiceName
```
ygalblum, line 93: Why is it a `var`? Do we want to be able to change it? If not, you can just expose the const by renaming it

main.go
```diff
@@ -18,7 +18,14 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"github.com/project-flotta/osbuild-operator/internal/composer"
```
ygalblum, line 24: This import should be at the last section

main.go
```diff
@@ -185,3 +206,34 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func createClient() (*http.Client, error) {
+	certDirFormat := "/etc/osbuild/certs/%s"
```
ygalblum, line 211: You should use `path.Join` instead of `Sprintf`

ygalblum, line 211: Consider moving the certs directory string to consts

main.go
```diff
@@ -185,3 +206,34 @@ func main() {
...
+	caCertPool := x509.NewCertPool()
+	caCertPool.AppendCertsFromPEM(caCert)
+
+	tlsConfig = &tls.Config{
+		MinVersion:   tls.VersionTLS12,
+		RootCAs:      caCertPool,
+		Certificates: []tls.Certificate{cert},
+	}
+
+	tlsConfig.BuildNameToCertificate()
```
ygalblum, line 236: I think this call is deprecated: https://pkg.go.dev/crypto/tls#Config.BuildNameToCertificate

ygalblum (changes requested):

controllers/osbuild_controller.go
ygalblum: Sure. You should retry, but why wait 2 minutes?

controllers/osbuild_controller.go
bardielle: I totally changed it, can you r-review it now?

controllers/osbuild_controller.go
bardielle: Change it to 10 seconds

controllers/osbuildenvconfig_controller.go
```diff
@@ -87,7 +87,8 @@ const (
 	workerAPIInternalPort   = 18700
 	workerAPIExternalPort   = 8700
 
-	envoyProxyCertsDir = "/etc/certs"
+	envoyProxyCertsDir   = "/etc/certs"
+	EnvoyOSBuildCertsDir = "/etc/osbuild/certs"
```
ygalblum, line 91: First, this const does not belong here as it is not used by this controller. Second, the name is incorrect since it is used to find the Operator's certificates inside the Operator's container. So, its names should not have `Envoy` in it (same as the strings for the worker certificate do not mention envoy)

main.go
```diff
@@ -185,3 +207,32 @@ func main() {
...
 	}
 }
+
+func createClient() (*http.Client, error) {
+	ca := path.Join(controllers.EnvoyOSBuildCertsDir, "ca.crt")
+	tlsCert := path.Join(controllers.EnvoyOSBuildCertsDir, "tls.crt")
+	tlsKey := path.Join(controllers.EnvoyOSBuildCertsDir, "tls.key")
+	var tlsConfig *tls.Config
+
+	caCert, err := ioutil.ReadFile(filepath.Clean(ca))
```
ygalblum, line 217: What do you need `Clean` for?

ygalblum (changes requested):

controllers/osbuild_controller.go
```diff
@@ -47,16 +82,254 @@ type OSBuildReconciler struct {
...
+		logger.Info("create an edge-container")
+		err = r.postComposeEdgeContainer(ctx, logger, osBuild)
+		if err != nil {
+			logger.Info("failed to create an edge-container")
+			return ctrl.Result{Requeue: true, RequeueAfter: requeueForLongDuration}, err
+		}
+
+		logger.Info("new job created for edge-container, requeue to sample its status")
+		return ctrl.Result{Requeue: true, RequeueAfter: requeueForLongDuration}, nil
+	} else if lastBuildStatus == startedContainerBuild {
```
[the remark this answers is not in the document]
bardielle: I don't think so: 
If in the current iteration  ContainerComposeId is not empty and the `lastBuildStatus` is finished we want to go to the next if-else (for building an edge installer job )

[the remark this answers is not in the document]
ygalblum: Still, whenever you have an `if` branch that always returns, there is no need for `else`. I didn't say they is no need for another `if`, just that there is no need for `else if`. The `else` protects your code from entering that branch after the `if` section is done. But, since you always return, the protection is not needed

controllers/osbuild_controller.go
```diff
@@ -47,16 +82,267 @@ type OSBuildReconciler struct {
...
+		Architecture:  string(targetImage.Architecture),
+		ImageType:     composer.ImageTypes(targetImageType),
+		Ostree:        (*composer.OSTree)(targetImage.OSTree.DeepCopy()),
+		UploadOptions: &uploadOptions,
+	}
+	return imageRequest
+}
+
+func (r *OSBuildReconciler) createCustomizations(osbuildCustomizations *osbuildv1alpha1.Customizations) composer.Customizations {
+	if &osbuildCustomizations == nil {
```
ygalblum, line 313: You should check `osbuildCustomizations == nil`, it's already the pointer you need to check

main.go
```diff
@@ -51,6 +59,10 @@ import (
 	//+kubebuilder:scaffold:imports
 )
 
+const (
+	OSBuildCertsDir = "/etc/osbuild/certs"
```
ygalblum, line 63: Should this be exported?

ygalblum (changes requested):

controllers/osbuild_controller.go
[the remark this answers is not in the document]
bardielle: got it now

main.go
bardielle: not anymore

controllers/osbuild_controller.go
```diff
@@ -47,16 +82,305 @@ type OSBuildReconciler struct {
...
+		}
 
-	// TODO(user): your logic here
+		// if the build was finished successfully and the target image is edge-container than return
+		if composeStatus == composer.ComposeStatusValueSuccess && osBuild.Spec.Details.TargetImage.TargetImageType == edgeContainerImgType {
+			logger.Info(fmt.Sprintf("the job ID %s, Finished", osBuild.Status.ContainerComposeId))
+			return ctrl.Result{}, nil
+		}
+	}
+
```
ygalblum, line 150: I think we're missing a check here. When building container is done (successfully or not), you'll get another reconcile call. This time lastBuildStatus will not be `startedContainerBuild` but either `containerBuildDone` or `failedContainerBuild` (depending on success of failure of the build). 
Then, if it failed, or succeeded, but the target is Container, you need to complete the process.
I think you'll get the cleanest code if you always requeue at the end of the previous `if` statement. This way, once the build is done, the Reconcile method we'll be called again, the status will be fetched again.

controllers/osbuild_controller.go
```diff
@@ -47,16 +82,305 @@ type OSBuildReconciler struct {
...
+		if composeStatus == composer.ComposeStatusValuePending {
+			logger.Info(fmt.Sprintf("the job ID %s, is still in progress", osBuild.Status.ContainerComposeId))
+			return ctrl.Result{Requeue: true, RequeueAfter: requeueForLongDuration}, nil
+		}
+
+		// the build was failed - return with error
+		if composeStatus == composer.ComposeStatusValueFailure {
+			err = fmt.Errorf("failed to build edge container")
+			logger.Error(err, "")
+			return ctrl.Result{}, err
```
ygalblum, line 141: From the investigation you did in the RequeueAfter, I now understand that if Reconcile returns an error, it will be called again. Do we want to be called again?
We haven't discussed emitting K8S events, but maybe we should emit an event here and return without an error

controllers/osbuild_controller.go
```diff
@@ -47,16 +82,305 @@ type OSBuildReconciler struct {
...
+		return err
+	}
+
+	composerResponse, err := composer.ParsePostComposeResponse(response)
+	if err != nil {
+		logger.Error(err, "failed parsing the response of postCompose")
+		return err
+	}
+	if composerResponse.StatusCode() != http.StatusCreated {
+		errorMsg := fmt.Sprintf("postCompose request was failed for OSBuild %s, with status code %v", osBuild.Name, composerResponse.StatusCode())
```
ygalblum, line 266: Rephrase, remove `was`

controllers/osbuild_controller.go
```diff
@@ -47,16 +82,305 @@ type OSBuildReconciler struct {
...
+		err = fmt.Errorf(errorMsg)
+		logger.Error(err, "postCompose request was failed")
```
ygalblum, line 268: Rephrase, remove `was`

controllers/osbuild_controller.go
```diff
@@ -47,16 +82,305 @@ type OSBuildReconciler struct {
...
+
+	if osBuild.Status.Conditions == nil {
+		osBuild.Status.Conditions = []osbuildv1alpha1.OSBuildCondition{}
+	}
+
+	conditionArrLen := len(osBuild.Status.Conditions)
+	if conditionArrLen > 0 {
+		lastConditionType := osBuild.Status.Conditions[conditionArrLen-1].Type
+		if lastConditionType == typeCondition {
+			logger.Info("The conditionType was left the same as before ", " lastConditionType ", lastConditionType)
```
ygalblum, line 301: Rephrase - `conditionType did not change`

controllers/osbuild_controller.go
```diff
@@ -47,16 +82,305 @@ type OSBuildReconciler struct {
...
+	}
+
+	containerComposeId := composerResponse.JSON201.Id.String()
+	logger.Info("postComposer request was sent and trigger a new compose ID %s", containerComposeId)
+
+	return r.updateOSBuildStatus(ctx, logger, osBuild, edgeContainerJobStillRunningMsg, startedContainerBuild, containerComposeId, emptyComposeID)
+}
+
+func (r *OSBuildReconciler) updateOSBuildStatus(ctx context.Context, logger logr.Logger, osBuild *osbuildv1alpha1.OSBuild,
+	msg string, typeCondition osbuildv1alpha1.OSBuildConditionType, containerComposeId string, isoComposeId string) error {
```
ygalblum, line 283: Sorry for nitpicking, but can you rename `typeCondition` to `conditionType`?

main.go
```diff
@@ -134,9 +144,23 @@ func main() {
 		}
 	}
 
+	setupLog.Info("Create a composer client")
+	httpClient, err := createClient()
```
ygalblum, line 148: Sorry for raising it only now, but if we already have a method that creates a client, why not have it return the `composerClient`

main.go
```diff
@@ -134,9 +144,23 @@ func main() {
...
+	if err != nil {
+		setupLog.Error(err, "unable to create http client")
+		os.Exit(1)
+	}
+
+	composerClient := composer.Client{
+		Server:         fmt.Sprintf("https://%s/api/image-builder-composer/v2/", controllers.ComposerComposerAPIServiceName),
```
ygalblum, line 155: Consider moving the format string to a const

ygalblum (changes requested):

controllers/osbuild_controller.go
bardielle: It sounds the right thing - any how I think it's better to return nil instead of the error

controllers/osbuild_controller.go
```diff
@@ -47,16 +82,336 @@ type OSBuildReconciler struct {
...
+		if osbuildCustomizations.Services.Disabled != nil {
+			services.Disabled = &osbuildCustomizations.Services.Disabled
+		}
+		composerCustomizations.Services = &services
+	}
+	if osbuildCustomizations.Packages != nil {
+		composerCustomizations.Packages = &osbuildCustomizations.Packages
+	}
+
+	return &composerCustomizations
```
ygalblum, line 409: You have the same issue with the Customization structure as a whole since it is not mandatory either. So, if Customizations is empty (meaning all the `if` statements were `false`), you should return `nil`

ygalblum (changes requested):

ygalblum (changes requested):

PR #110, opened 2022-07-17
WIP: Load repositories from files

ygalblum:
Create a ConfigMap for the repositories jsons for all distros
Attach the ConfigMap to the manager
Get the repositories information from the stored files


bardielle (approved):
Tested it and it worked

PR #114, opened 2022-07-18
OSBuildConfig - Add Repositories list

ygalblum:
CRD
---
Add optional Repositories field to the TargetImage struct
Add Repository type

osbuild_controller
------------------
Remove reading the repositories files from the OSBuild Controller

Config
------
Add RepositoriesDir to the configuration to allow overriding

osbuild manifest
----------------
Read the repositories configuration from local files
Merge the local configuration with the OSBuildConfig specific ones if provided
Adjust UT

Signed-off-by: Ygal Blum <ygal.blum@gmail.com>


api/v1alpha1/osbuildconfig_types.go
```diff
@@ -97,6 +97,8 @@ type TargetImage struct {
 	TargetImageType TargetImageType `json:"targetImageType"`
 	// OSTree is the OSTree configuration of the build (optional)
 	OSTree *OSTreeConfig `json:"osTree,omitempty"`
+	// Repositories is the list of additional custom RPM repositories to use when building the image (optional)
+	Repositories *[]Repository `json:"repositorys,omitempty"`
```
bardielle, line 101: I think `CustomRepositories` is a better name

ygalblum: For the type I disagree. These are Repositories nothing Custom about them.
As for the name, I try to stick as close as possible to the OSBuild terminology. As such and since this field translates to the `Repositories`, I think it should also be names `Repositories`

internal/manifests/osbuild.go
```diff
@@ -119,6 +129,45 @@ func (o *OSBuildCreator) Create(ctx context.Context, osBuildConfig *osbuildv1alp
...
 }
 
+func mergeRepositories(osBuildConfigSpecDetails *osbuildv1alpha1.BuildDetails) error {
+	repos, err := getDefaultRepositories(osBuildConfigSpecDetails.Distribution, osBuildConfigSpecDetails.TargetImage.Architecture)
+	if err != nil {
+		return nil
+	}
+
+	if osBuildConfigSpecDetails.TargetImage.Repositories != nil {
+		repos = append(repos, *osBuildConfigSpecDetails.TargetImage.Repositories...)
```
bardielle, line 139: what about the case that would be a conclusion between the default repos to the customize repos?

bardielle (approved):

PR #115, opened 2022-07-18
Add build URL to osbuild status

bardielle:


controllers/osbuild_controller.go
```diff
@@ -197,12 +198,25 @@ func (r *OSBuildReconciler) updateContainerComposeStatus(ctx context.Context, lo
...
 		logger.Error(err, "failed to update OSBuild condition status")
 		return "", err
 	}
-	return composeStatus, nil
+	return status, nil
+}
+
+func (r *OSBuildReconciler) getBuildUrl(composeStatus *composer.ComposeStatus) string {
+	buildUrl := emptyURL
+	imageStatuses := *(*composeStatus).ImageStatuses
```
ygalblum, line 214: Do you need to dereference `composeStatus`? Wouldn't `*composeStatus.ImageStatuses` work? In addition you should check that it is not `nil` before dereferencing

controllers/osbuild_controller.go
```diff
@@ -197,12 +198,25 @@ func (r *OSBuildReconciler) updateContainerComposeStatus(ctx context.Context, lo
...
+	if len(imageStatuses) > 0 {
+		buildUrl = imageStatuses[0].UploadStatus.Options.(*composer.AWSS3UploadStatus).Url
```
ygalblum, line 216: You should validate that the casting didn't fail:
```
uploadStatus, ok := imageStatuses[0].UploadStatus.Options.(*composer.AWSS3UploadStatus)
if !ok {
  return error
}
```

ygalblum (approved):

PR #120, opened 2022-07-20
Add support for external osbuild-worker machine

ygalblum:
main
----
Add Route to the Scheme
Create a RouteRepository and pass to the controller

controller
----------
Expose the Composer's Worker API endpoint as a Route
Add the Route's host to the Composer's certificate
Make the playbook worker specific because the endpoint address may be different
Pass the endpoint address (service or route) to the VM via the playbook

playbook
--------
Force registration to RH because the module is not idempotant

rbac
----
Add permissions for Route

repository
----------
Create a repository for Route

Unit Tests
----------
Adjust to the updated flow

Sample and README
-----------------
Add yaml files and explanation for external VM using CNV


README.md
```diff
@@ -112,10 +112,34 @@ Currently the controller does not support creating the PSQL server on its own, m
   oc create -f config/creating_env/postgress_secret.yaml
   ```
 
+## External VM using CNV
```
bardielle, line 115: External VM using CNV -> External VM *Worker* using CNV

bardielle: I think it's an assumption that we have CNV - but we didn't add it as a prerequisite

config/creating_env/external-worker-vm.yaml
```diff
@@ -0,0 +1,95 @@
...
+kind: VirtualMachine
+metadata:
+  annotations:
+    description: OSBuild Worker
+    name.os.template.kubevirt.io/rhel8.5: "Red Hat Enterprise Linux 8.0 or higher"
+  labels:
+    app: osbuild-worker
+    workload.template.kubevirt.io/server: "true"
+    os.template.kubevirt.io/rhel8.5: "true"
+  namespace: osbuild
```
bardielle, line 11: Should it be run on the same namespace? 
If not - please remove it and add the "-n" flag to the cmd

controllers/osbuildenvconfig_controller_test.go
```diff
@@ -322,416 +341,359 @@ var _ = Describe("OSBuildEnvConfig Controller", func() {
...
+						composerComposerAPIServiceName = "osbuild-composer"
+					)
+					var (
+						composerCertificate certmanagerv1.Certificate
 					)
+
 					BeforeEach(func() {
-						var sslMode osbuildv1alpha1.DBSSLMode = "disable"
+						composerWorkerAPIRoute.Status = routev1.RouteStatus{
+							Ingress: []routev1.RouteIngress{
```
bardielle, line 461: Maybe the RouteStatus can be part of the on-top beforeEach func and just edit the "status"

controllers/osbuildenvconfig_controller_test.go
```diff
@@ -893,26 +899,81 @@ var _ = Describe("OSBuildEnvConfig Controller", func() {
 													})
 												})
 
-												Context("ConfigMap for the osbuild-worker config exists", func() {
+												Context("ConfigMap for the ansible config exists", func() {
```
bardielle, line 902: I know that we already talked about that - but it become difficult to track the test cases with that adaptation

README.md
ygalblum: Regarding the second comment. CNV is a requirement for local workers. For the external worker, this in only a sample. In practice, there is no point in using CNV based VMs as external workers since one can achieve the same with internal workers

controllers/osbuildenvconfig_controller_test.go
ygalblum: Not sure what you're referring to.
The root cause of this complicated unit test is the fact that we have to test the main function (Reconcile) as a whole. We cannot test the internal functions since the test is in a different package, and we can't mock the internal methods to contain the tests of the external one.
Furthermore, with all the contexts I defined, when a test fails, it's really easy to understand where in the flow you are.

README.md
bardielle: I meant - CNV operator is required for the workers that created by the osbuild operator

controllers/osbuildenvconfig_controller_test.go
ygalblum: What's the benefit in that?

README.md
ygalblum: We can add a line about it in the `Requirements` section. But, it how does it relate to this PR?

bardielle: Not specific related but I think it can be part of that PR

controllers/osbuildenvconfig_controller_test.go
bardielle: less duplicated code

README.md
ygalblum: Still think this should be done separately, but done

bardielle (approved):

PR #129, opened 2022-07-26
Make sure the certificate secrets are owned so they will deleted with the instance

ygalblum:
Instead of deleting the secrets upon the deletion of the CR, make the CR
their owner so that K8S deletes them automatically

### DIFF
--- a/controllers/osbuildenvconfig_controller.go
+++ b/controllers/osbuildenvconfig_controller.go
@@ -394,6 +394,14 @@ func (r *OSBuildEnvConfigReconciler) ensureComposerExists(ctx context.Context, r
 		return true, nil
 	}
 
+	updated, err := r.ensureSecretOwnedByInstance(ctx, reqLogger, instance, composerCertificateName)
+	if err != nil {
+		return false, err
+	} else if updated {
+		reqLogger.Info("Updated owner reference for the composer certificate secret", "secretName", composerCertificateName)
+		return true, nil
+	}
+
 	composerDeploymentParams := composerDeploymentParameters{
 		ComposerDeploymentNamespace:      conf.GlobalConf.WorkingNamespace,
 		ComposerDeploymentName:           composerDeploymentName,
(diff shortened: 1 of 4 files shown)


controllers/osbuildenvconfig_controller.go
```diff
@@ -634,6 +650,36 @@ func (r *OSBuildEnvConfigReconciler) generateCertificate(ctx context.Context, in
...
+
+func (r *OSBuildEnvConfigReconciler) ensureSecretOwnedByInstance(ctx context.Context, reqLogger logr.Logger, instance *osbuildv1alpha1.OSBuildEnvConfig, secretName string) (bool, error) {
+	secret, err := r.SecretRepository.Read(ctx, secretName, conf.GlobalConf.WorkingNamespace)
+	if err != nil {
+		return false, err
+	}
+
+	if len(secret.ObjectMeta.OwnerReferences) > 0 {
+		// Do not check who is the owner since the cert-manager may be configured to own its secrets
+		// In that case, the secret will be deleted with the certificate and the osbuildenvconfig doesn't need to own it
```
bardielle, line 665: :+1:

controllers/osbuildenvconfig_controller_test.go
```diff
@@ -886,40 +860,31 @@ var _ = Describe("OSBuildEnvConfig Controller", func() {
...
-													)
-
-													var (
-														workerConfigAnsibleConfigConfigMap = corev1.ConfigMap{
-															ObjectMeta: metav1.ObjectMeta{
-																Namespace: operatorNamespace,
-																Name:      workerConfigAnsibleConfigConfigMapName,
-															},
-														}
+														workerConfigAnsibleConfigConfigMapName = "osbuild-worker-setup-ansible-config"
```
bardielle, line 865: Why not to use the const from the controller?

ygalblum: I can't use the const from the controller because the test is in a different package so it does not have access to internal variables and consts. I could expose them, but, I think it's bad practice.
Let's say someone accidentally changes the value of a const (typing a key without noticing where the cursor is), using the same const will not catch it, while separately defining the expected value will.

bardielle: Yes I meant to expose them but if you prefer not to I get it.

bardielle (approved):

PR #136, opened 2022-07-31
Add unittests for internal and external worker

ygalblum:
Add interface and mock to the SSH package


controllers/osbuildenvconfig_controller_test.go
```diff
@@ -69,9 +83,12 @@ var _ = Describe("OSBuildEnvConfig Controller", func() {
 		certificateRepository      *certificate.MockRepository
 		configMapRepository        *configmap.MockRepository
 		deploymentRepository       *deployment.MockRepository
+		jobRepository              *job.MockRepository
 		routeRepository            *route.MockRepository
 		serviceRepository          *service.MockRepository
 		secretRepository           *secret.MockRepository
+		virtualmachineRepository   *virtualmachine.MockRepository
```
bardielle, line 90: virtualmachineRepository -> virtualMachineRepository

controllers/osbuildenvconfig_controller_test.go
```diff
@@ -69,9 +83,12 @@ var _ = Describe("OSBuildEnvConfig Controller", func() {
...
+		sshkeyGenerator            *sshkey.MockSSHKeyGenerator
```
bardielle, line 91: sshkeyGenerator -> sshKeyGenerator

controllers/osbuildenvconfig_controller_test.go
```diff
@@ -961,25 +1006,639 @@ var _ = Describe("OSBuildEnvConfig Controller", func() {
...
+															It("Should return an error if failed to get the secret for the osbuild-worker ssh keys", func() {
+																// given
+																secretRepository.EXPECT().Read(requestContext, workerSSHKeysSecretName, operatorNamespace).Return(nil, errFailed)
 																// when
 																result, err := reconciler.Reconcile(requestContext, request)
 																// then
-																Expect(err).To(BeNil())
-																Expect(result).To(Equal(resultDone))
+																Expect(err).To(Equal(errFailed))
+																Expect(result).To(Equal(resultRequeue))
```
bardielle, line 1052: I know it was merged but I think it needs to be changed (if not in that PR  in a new one) -  there is no reason to return a result like:   `return ctrl.Result{Requeue: true}, err` because as you can see here - https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L322 the error has been validated first

controllers/osbuildenvconfig_controller_test.go
```diff
@@ -961,25 +1006,639 @@ var _ = Describe("OSBuildEnvConfig Controller", func() {
...
+																	})
+																})
+															})
+
+															Context("Secret for osbuild-worker ssh keys exists", func() {
+																BeforeEach(func() {
+																	secretRepository.EXPECT().Read(requestContext, workerSSHKeysSecretName, operatorNamespace).Return(workerSSHKeysSecret, nil)
+																})
+
+																It("Should return an error if failed to get the virtualmachine for the internal builder", func() {
```
bardielle, line 1103: virtualmachin -> VirtualMachine  (just because I saw in other places you used the upper case for the description )

controllers/osbuildenvconfig_controller_test.go
```diff
@@ -961,25 +1006,639 @@ var _ = Describe("OSBuildEnvConfig Controller", func() {
...
+
+																		var (
+																			internalBuilderCertificateName = fmt.Sprintf(workerCertificateNameFormat, internalBuilderName)
+																			internalBuilderCertificate     *certmanagerv1.Certificate
+																		)
+
+																		BeforeEach(func() {
+																			serviceRepository.EXPECT().Read(requestContext, workerSSHServiceName, operatorNamespace).Return(workerSSHService, nil)
+																			internalBuilderCertificate = &certmanagerv1.Certificate{
+																				ObjectMeta: metav1.ObjectMeta{
```
bardielle, line 1232: Maybe is can be part of the beforeEach section and be reused for line 1563

controllers/osbuildenvconfig_controller_test.go
```diff
@@ -961,25 +1006,639 @@ var _ = Describe("OSBuildEnvConfig Controller", func() {
...
+																											ObjectMeta: metav1.ObjectMeta{
+																												Namespace:       operatorNamespace,
+																												Name:            internalBuilderCertificateName,
+																												OwnerReferences: []metav1.OwnerReference{ownerReference},
+																											},
+																										}
+																										secretRepository.EXPECT().Read(requestContext, externalBuilderCertificateName, operatorNamespace).Return(externalBuilderCertificateSecret, nil)
+																									})
+
+																									Context("The configmap for the playbook of the external builder setup exists", func() {
```
bardielle, line 1574: configmap -> configMap

controllers/osbuildenvconfig_controller_test.go
```diff
@@ -961,25 +1006,639 @@ var _ = Describe("OSBuildEnvConfig Controller", func() {
...
+																													Namespace: operatorNamespace,
+																												},
+																											}
+																										)
+
+																										BeforeEach(func() {
+																											configMapRepository.EXPECT().Read(requestContext, externalBuilderSetupPlaybookConfigMapName, operatorNamespace).Return(externalBuilderSetupPlaybookConfigMap, nil)
+																										})
+
+																										Context("The configmap for the inventory of the external builder setup exists", func() {
```
bardielle, line 1589: why did you split that context from the one in line 1574? 
It seems that it can be a reunion

controllers/osbuildenvconfig_controller_test.go
ygalblum: Yes, you're right about the return value. We need to decide what's the right return value here. Should it return error at all? In any case, I don't think it should be part of this PR

controllers/osbuildenvconfig_controller_test.go
ygalblum: Not sure what you mean. First, this is a BeforeEach and second this is a certificate and the code in line 1563 is for a secret

bardielle: the ObjectMeta looks the same

controllers/osbuildenvconfig_controller_test.go
bardielle: please open a ticket

controllers/osbuildenvconfig_controller_test.go
ygalblum: All the contexts from line 1574 till the end basically duplicate what was already tested for the Internal Worker, but this time for the External one. Theoretically, they can all be consolidated into one. But, I split them into multiple ones for two reasons:
1. We can still add tests in them
2. It makes the test log more readable

controllers/osbuildenvconfig_controller_test.go
ygalblum: Done

controllers/osbuildenvconfig_controller_test.go
ygalblum: Done

controllers/osbuildenvconfig_controller_test.go
ygalblum: Done

controllers/osbuildenvconfig_controller_test.go
ygalblum: Done (and I found more cases of this fix)

controllers/osbuildenvconfig_controller_test.go
ygalblum: But, one is for the Certificate and the other is for the Secret. It's just that we configure the certificate to create a secret with the same name. I don't think consolidating it will make the code more readable.

controllers/osbuildenvconfig_controller_test.go
ygalblum: Done

controllers/osbuildenvconfig_controller_test.go
bardielle: OK :+1:

bardielle (approved):

PR #138, opened 2022-08-01
Wait for the internal worker VM to be ready before scheduling a job

ygalblum:
Adjust UT


ygalblum:
Something in the diff that GH shows for the UT went really wrong. The change adds a test for this step and the rest is one indentation to the right. Consider downloading the branch and looking at the diff locally

bardielle (approved):

PR #143, opened 2022-08-07
Change osbuild CRD to contains two different build types

bardielle:


internal/manifests/osbuild.go
```diff
@@ -184,9 +190,7 @@ func (o *OSBuildCreator) applyTemplate(ctx context.Context, osBuildConfig *osbui
 		if err != nil {
 			return nil, nil, err
 		}
-		if kickstartConfigMap != nil {
```
ygalblum, line 187: Why did you remove this case? The configMap is still being created

ygalblum (changes requested):

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -104,9 +115,13 @@ type TargetImage struct {
 // +kubebuilder:validation:Enum=x86_64;aarch64
 type Architecture string
 
-// +kubebuilder:validation:Enum=edge-container;edge-installer
 type TargetImageType string
```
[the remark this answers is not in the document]
bardielle: I was sure that by adding the const section it will keep the validation - I added it again

internal/manifests/osbuild.go
bardielle: because kicktsart now is part of edgeinstallerdetals and it cannot be set automatically -this section was moved to the create function as part of `EdgeInstallerImageType`

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -63,6 +63,17 @@ type BuildDetails struct {
 	TargetImage TargetImage `json:"targetImage"`
 }
 
+// EdgeInstallerBuildDetails includes all the information needed to build the edge-installer image
+type EdgeInstallerBuildDetails struct {
+	// Distribution is the name of the O/S distribution
+	Distribution string `json:"distribution"`
+	// OSTree is the OSTree configuration of the build (optional)
+	OSTree *OSTreeConfig `json:"osTree,omitempty"`
```
ygalblum, line 71: I think that for the EdgeInstaller the OSTreeConfig is mandatory

internal/manifests/osbuild.go
ygalblum: If that's so, then we need to move the creation of the ConfigMap as well.

ygalblum (changes requested):

api/v1alpha1/osbuildconfig_types.go
bardielle: I'm not sure about that - I looked into the handler code and I didn't find such a validation 
but I think we can require it from the user

internal/manifests/osbuild.go
bardielle: I moved the kickstartConfigMap creation to the EdgeInstallerImageType case, but keep the setting owner in the full flow (I mean [here:](https://github.com/project-flotta/osbuild-operator/pull/143/files#diff-eb685724eddffc42344c24fdd392f723af5051deda96dc5cb59dbfc4f69ff9c2L119) ) because only if the osBuild was created I would like to set the owner.

bardielle: Done - Please resolve if it's fine by you

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -63,6 +63,17 @@ type BuildDetails struct {
 	TargetImage TargetImage `json:"targetImage"`
 }
 
+// EdgeInstallerBuildDetails includes all the information needed to build the edge-installer image
+type EdgeInstallerBuildDetails struct {
```
ygalblum, line 67: Since this type is only used in the OSBuildSpec, its definition should move to the osbuild_types.go file

internal/manifests/osbuild.go
ygalblum: Not sure if it works or not. I guess we'll need to implement the edge-installer and the template to find out. Resolving for now

ygalblum (changes requested):

api/v1alpha1/osbuildconfig_types.go
ygalblum: First, it's not from the user since this struct is only in the OSBuild which is created by the controller. Second, the reason it is not mandatory in the handler is that they use default values if not provided. But, in this specific case, the idea is to build an edge-installer from a previously built edge-container. So, we must provide these details.

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -53,7 +53,7 @@ type ParameterValue struct {
 	Value string `json:"value"`
 }
 
-// BuildDetails includes all the information needed to build the image
+// BuildDetails includes all the information needed to build the edge-container image
```
ygalblum, line 56: As far as the user who creates the OSBuildConfig CR is concerned BuildDetails is the same for all image types. It's not specific to edge-container. Please revert

ygalblum (changes requested):

ygalblum (approved):

PR #144, opened 2022-08-07
Get image name into env using kustomize vars

ygalblum:
It appears that you cannot extract the container's image name using valueFrom. So, instead use Kustomize variable to align the value of the environment variable and the container's image

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -53,3 +53,13 @@ configMapGenerator:
 - name: default-repos-rhel-91
   files:
   - repositories/rhel-91.json
+
+vars:
+  - name: MANAGER_CONTAINER_IMAGE
+    objref:
+      group: apps
+      kind: Deployment
+      name: controller-manager
+      version: v1
+    fieldref:
+      fieldpath: spec.template.spec.containers.[name=manager].image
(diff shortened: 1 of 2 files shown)


config/manager/kustomization.yaml
```diff
@@ -53,3 +53,13 @@ configMapGenerator:
 - name: default-repos-rhel-91
   files:
   - repositories/rhel-91.json
+
+vars:
```
bardielle, line 57: There is a default value?

ygalblum: What do you mean? It gets the value from the image field of the manager container in the deployment spec. Why would it have a default value?

bardielle: I thought the user setting it as part of the deployment

ygalblum: Again, I don't understand what you mean. The code needs to know the container image to use for the ISO repackaging job. For that, we added a new configuration value that gets its value from the Environment Variables. We agreed that we will put the executable inside the same container as the manager and httpapi. So, we don't want to hard-code this value because during development the value changes. But, we don't want to change it manually. Instead, I'm using Kustomize's `vars` mechanism

bardielle (approved):

PR #150, opened 2022-08-09
OSBuild Controller - fix condition names

ygalblum:
Change UT to use local variables to compare between expected and returned

### DIFF
--- a/controllers/osbuild_controller.go
+++ b/controllers/osbuild_controller.go
@@ -46,9 +46,9 @@ const (
 	EdgeContainerJobStillRunningMsg = "Edge-container job is still running"
 
 	// OSBuildConditionTypes values
-	ContainerBuildDone    = "ContainerBuildDone"
-	FailedContainerBuild  = "FailedContainerBuild"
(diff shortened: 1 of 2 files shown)


controllers/osbuild_controller.go
```diff
@@ -46,9 +46,9 @@ const (
 	EdgeContainerJobStillRunningMsg = "Edge-container job is still running"
 
 	// OSBuildConditionTypes values
-	ContainerBuildDone    = "ContainerBuildDone"
```
bardielle, line 49: In my next PR I'm changing those consts to be in v1alpha1 as I saw it was done in [kubevirt](https://github.com/kubevirt/kubevirt/blob/f3ddc0c65b54f4d1d0ad9d583f330cdff7385dd3/staging/src/kubevirt.io/api/core/v1/types.go#L1860)... 
In their case - the values are in the Upper case

ygalblum: Currently it's broken and builds cannot be triggered. So, let's merge this PR and change it later after testing

bardielle (approved):

PR #158, opened 2022-08-17
Change OSBuildConfig controller to create seperated OSBuild instance …

bardielle:
…per target image type


internal/osbuildconfig/osbuildconfig_handler.go
```diff
@@ -7,30 +7,35 @@ package osbuildconfig
 
 import (
 	"fmt"
+	osbuildv1alpha1 "github.com/project-flotta/osbuild-operator/api/v1alpha1"
```
ygalblum, line 10: Please move the imports to their corresponding location

ygalblum (changes requested):

api/v1alpha1/osbuildconfig_types.go
```diff
@@ -94,7 +94,7 @@ type TargetImage struct {
 	// Architecture defines target architecture of the image
 	Architecture Architecture `json:"architecture"`
 	// TargetImageType defines the target image type
-	// +kubebuilder:validation:Enum=edge-container;edge-installer
+	// +kubebuilder:validation:Enum=edge-container;edge-installer;qcow2
```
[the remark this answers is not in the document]
bardielle: I will open a separate PR for qcow2

api/v1alpha1/osbuildconfig_webhook.go
```diff
@@ -67,6 +68,13 @@ func (r *OSBuildConfig) ValidateUpdate(old runtime.Object) error {
 
 	err := fmt.Errorf("cannot update immutable fields")
 
+	for _, c := range r.Status.Conditions {
```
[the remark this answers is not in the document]
bardielle: I change the logic here
please re-review it

controllers/osbuild_controller.go
```diff
@@ -99,60 +92,46 @@ func (r *OSBuildReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-		return r.buildEdgeInstallerImage(ctx, logger, osBuild)
-	}
-
-	return ctrl.Result{}, nil
-}
-
-func (r *OSBuildReconciler) buildEdgeContainerImage(ctx context.Context, logger logr.Logger, osBuild *osbuildv1alpha1.OSBuild) (ctrl.Result, error) {
-	if osBuild.Status.ContainerComposeId == EmptyComposeID {
+	if osBuild.Status.ComposeId == EmptyComposeID {
 		// if the edge container wasn't created yet - schedule a new build
```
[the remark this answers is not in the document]
bardielle: Done

controllers/osbuild_controller.go
```diff
@@ -99,60 +92,46 @@ func (r *OSBuildReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-		return ctrl.Result{Requeue: true, RequeueAfter: RequeueForLongDuration}, nil
+		logger.Info("create a new image")
+		return r.postComposeNewImage(ctx, logger, osBuild)
 	}
 
-	lastBuildStatus := osbuildv1alpha1.OSBuildConditionType("")
-	if osBuild.Status.Conditions != nil {
-		conditionLen := len(osBuild.Status.Conditions)
-		lastBuildStatus = osBuild.Status.Conditions[conditionLen-1].Type
+	lastBuildStatus := osbuildv1alpha1.ConditionType("")
```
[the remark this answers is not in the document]
bardielle: Done

controllers/osbuild_controller.go
```diff
@@ -99,60 +92,46 @@ func (r *OSBuildReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 
-	case failedContainerBuild:
+	case osbuildv1alpha1.ConditionFailed:
 		logger.Error(fmt.Errorf("failed to build edge container"), "")
 		return ctrl.Result{}, nil
 
-	case containerBuildDone:
-		logger.Info(fmt.Sprintf("the job ID %s, Finished", osBuild.Status.ContainerComposeId))
+	case osbuildv1alpha1.ConditionReady:
+		//	in case target image is edge-installer - the build was finished successfully - continue with repackaging the iso image
```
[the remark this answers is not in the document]
bardielle: I moved it to the `Processing` case

controllers/osbuild_controller.go
```diff
@@ -161,55 +140,28 @@ func (r *OSBuildReconciler) buildEdgeContainerImage(ctx context.Context, logger
...
-		}
-
-		return ctrl.Result{Requeue: true}, nil
-
-	case failedIsoBuild:
-		// the build was failed - return with error
-		logger.Error(fmt.Errorf("failed building the edge installer"), "")
-		return ctrl.Result{}, nil
+func (r *OSBuildReconciler) initConditionArray(ctx context.Context, logger logr.Logger, osBuild *osbuildv1alpha1.OSBuild) {
+	osBuild.Status.Conditions = append(osBuild.Status.Conditions, osbuildv1alpha1.Condition{
```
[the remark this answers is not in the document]
bardielle: Done

controllers/osbuild_controller.go
```diff
@@ -248,51 +200,27 @@ func (r *OSBuildReconciler) getBuildUrl(logger logr.Logger, composeStatus *compo
...
 	if composeStatus == composer.ComposeStatusValuePending {
-		return r.updateOSBuildStatus(ctx, logger, osBuild, EdgeContainerJobStillRunningMsg, buildStartedValue, EmptyComposeID, EmptyComposeID, edgeContainerUrl, edgeInstallerUrl)
+		return r.updateOSBuildStatus(ctx, logger, osBuild, buildJobStillRunningMsg, osbuildv1alpha1.ConditionProgressing, EmptyComposeID, accessUrl)
 	}
 
 	return nil
 }
 
-func (r *OSBuildReconciler) postComposeEdgeContainer(ctx context.Context, logger logr.Logger, osBuild *osbuildv1alpha1.OSBuild) error {
+func (r *OSBuildReconciler) postComposeNewImage(ctx context.Context, logger logr.Logger, osBuild *osbuildv1alpha1.OSBuild) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
bardielle: I thought to handle it with  a different function and add an `if` in line 106 
I added a TODO comment

controllers/osbuild_controller.go
```diff
@@ -305,69 +233,69 @@ func (r *OSBuildReconciler) postComposeEdgeContainer(ctx context.Context, logger
...
-	conditionArrLen := len(osBuild.Status.Conditions)
-	if conditionArrLen > 0 {
-		lastConditionType := osBuild.Status.Conditions[conditionArrLen-1].Type
-		if lastConditionType == conditionType {
-			logger.Info("conditionType did not change ", " lastConditionType ", lastConditionType)
-			return nil
+		r.initConditionArray(ctx, logger, osBuild)
+	}
+
+	conditionsArr := osBuild.Status.Conditions
```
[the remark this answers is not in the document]
bardielle: it will be by value and will not update the array ar all (I already tried it and failed)

controllers/osbuild_controller.go
```diff
@@ -305,69 +233,69 @@ func (r *OSBuildReconciler) postComposeEdgeContainer(ctx context.Context, logger
...
+	for i := range conditionsArr {
+		if conditionsArr[i].Type == newConditionStatus {
+			conditionsArr[i].Status = metav1.ConditionTrue
+			conditionsArr[i].Message = &msg
+			conditionsArr[i].LastTransitionTime = &metav1.Time{Time: time.Now()}
+		} else {
+			conditionsArr[i].Status = metav1.ConditionFalse
+			conditionsArr[i].Message = nil
```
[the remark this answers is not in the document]
bardielle: Yes. and also update the LastTransitionTime field

controllers/osbuildconfig_controller.go
```diff
@@ -33,11 +36,24 @@ import (
...
 type OSBuildConfigReconciler struct {
-	client.Client
 	OSBuildConfigRepository osbuildconfig.Repository
+	OSBuildRepository       osbuild.Repository
 	OSBuildCRCreator        manifests.OSBuildCRCreator
 }
 
+const (
+	// Conditions Messages
+	ProgressMessage = "Last OSBuild instance creation is in Progress"
```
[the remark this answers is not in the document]
bardielle: No need it anymore since I removed the condition array from osbuildconfig controller

internal/osbuildconfig/osbuildconfig_handler.go
bardielle: Done

controllers/osbuildconfig_controller_test.go
```diff
@@ -0,0 +1,575 @@
...
+			osBuildConfigRepository.EXPECT().PatchStatus(requestContext, osbuildConfigInstance, gomock.Any()).Return(errFailed)
+			// when
+			result, err := reconciler.Reconcile(requestContext, request)
+			// then
+			Expect(err).To(BeNil())
+			Expect(result).To(Equal(resultShortRequeue))
+		},
+			Entry("because of LastKnownUserConfiguration is nil", nil, map[string]string{}),
+			Entry("because of LastKnownUserConfiguration is different from the current userConfiguration", &osbuildv1alpha1.UserConfiguration{Customizations: &osbuildv1alpha1.Customizations{Packages: []string{"pkg1"}}}, map[string]string{}),
+			Entry("because of the webHookAnnotationKey was changed", &osbuildv1alpha1.UserConfiguration{Customizations: &osbuildv1alpha1.Customizations{Packages: []string{"pkg1", "pkg2"}}}, map[string]string{"last_webhook_trigger_ts": "1111"}),
```
ygalblum, line 203: Another case is when both changed in the same Reconcile

controllers/osbuildconfig_controller.go
```diff
@@ -70,15 +81,139 @@ func (r *OSBuildConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	return false, nil
+}
+
+func (r *OSBuildConfigReconciler) updateOSBuildConfigCurrentStatus(ctx context.Context, logger logr.Logger, osBuildConfig *osbuilderv1alpha1.OSBuildConfig) (ctrl.Result, error) {
+	osBuildName := fmt.Sprintf("%s-%d", osBuildConfig.Name, *osBuildConfig.Status.LastVersion)
+	osBuild, err := r.OSBuildRepository.Read(ctx, osBuildName, osBuildConfig.Namespace)
 	if err != nil {
 		if errors.IsNotFound(err) {
-			return ctrl.Result{RequeueAfter: 30 * time.Second}, err
+			// should create a new one from the same target image type
```
ygalblum, line 153: I don't understand this `if`. Why should this happen and why is OK that it happened?

controllers/osbuildconfig_controller_test.go
```diff
@@ -0,0 +1,575 @@
...
+					Expect(err).To(BeNil())
+					Expect(result).To(Equal(resultLongRequeue))
+				})
+
+			})
+		})
+
+	})
+
+	Context("OSBuildConfig status is not Progressing state", func() {
```
ygalblum, line 447: Sorry, but, I don't understand what's going on in this `Context`, not why these are the expected behaviors (e.g. a new instance is created when in Failed status, i.e. retry) or where it is implemented

ygalblum (changes requested):

controllers/osbuildconfig_controller.go
```diff
@@ -70,15 +81,139 @@ func (r *OSBuildConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			// start with edge-container OSBuild instance
+			targetBuild = osbuilderv1alpha1.EdgeContainerImageType
+		}
+		return r.createOSBuildInstance(ctx, logger, osBuildConfig, targetBuild)
+	} else {
+		return r.updateOSBuildConfigCurrentStatus(ctx, logger, osBuildConfig)
+	}
+}
+
+func (r *OSBuildConfigReconciler) checkIfNewOSBuildInstanceIsNeeded(ctx context.Context, osBuildConfig *osbuilderv1alpha1.OSBuildConfig, logger logr.Logger) (bool, error) {
```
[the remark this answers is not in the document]
bardielle: Not sure how it can happen - it means that 2 different calls would be... 
And 2 new instances would be created

controllers/osbuildconfig_controller_test.go
bardielle: As I wrote before - the trigger webhook cannot change the object... 
Not sure how it can happen

controllers/osbuildconfig_controller.go
```diff
@@ -70,15 +81,139 @@ func (r *OSBuildConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	userConfiguration := osbuilderv1alpha1.UserConfiguration{}
+	// TODO: sort before
+	if osBuildConfig.Spec.Details.Customizations != nil {
+		userConfiguration.Customizations = osBuildConfig.Spec.Details.Customizations
+	}
+	if osBuildConfig.Spec.Template != nil {
+		userConfiguration.Template = osBuildConfig.Spec.Template
+	}
+
+	if osBuildConfig.Status.LastKnownUserConfiguration == nil || !cmp.Equal(osBuildConfig.Status.LastKnownUserConfiguration, userConfiguration) {
```
ygalblum, line 112: You have a bug here. `LastKnownUserConfiguration` is a pointer while `userConfiguration` is a struct. You need to pass `&userConfiguration`

ygalblum (changes requested):

controllers/osbuildconfig_controller.go
bardielle: For example - 
1- Create osbuildConfig
2- update it immediately so the OSBuild instance wasn't created yet 
in that case, I think we would like to create a new instance

controllers/osbuildconfig_controller_test.go
bardielle: It's an old test - as I wrote -I need to fix the UT, I will push it today

controllers/osbuildconfig_controller.go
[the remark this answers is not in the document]
ygalblum: It can happen just because of the way event handling is implemented. Imaging this scenario:
1. The node on which the operator is running fails and shuts down.
2. While OCP is rescheduling the operator the user updates the configuration, but also some automated procedure calls the webhook
3. The operator comes back to life examines the CR and as a result treats the config change disregarding the webhook timestamp change
4. At an unknown time in the future the operator restarts for some reason
5. The timestamp is in the status is still different than the one in the annotation and as a result another build is scheduled
This scenario is more elaborate and includes a downtime of the operator and hence might seem as an overreach. But, the downtime only comes to show that multiple things can change on the CR until the operator handles the event.

Now, I don't understand you comment about "2 instances would be created". Instances of what? If OSBuildConfig, then why? If OSBuild, then no, only one should be created

controllers/osbuildconfig_controller.go
```diff
@@ -70,15 +82,137 @@ func (r *OSBuildConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		}
+		return r.createOSBuildInstance(ctx, logger, osBuildConfig, targetBuild)
+	} else {
+		return r.updateOSBuildConfigCurrentStatus(ctx, logger, osBuildConfig)
+	}
+}
+
+func (r *OSBuildConfigReconciler) checkIfNewOSBuildInstanceIsNeeded(ctx context.Context, osBuildConfig *osbuilderv1alpha1.OSBuildConfig, logger logr.Logger) (bool, error) {
+	userConfiguration := osbuilderv1alpha1.UserConfiguration{}
+	// TODO: sort before
```
ygalblum, line 105: I think this `TODO` must be implemented before the change is merged. Otherwise, we're won't get a consistent behavior

controllers/osbuildconfig_controller.go
ygalblum: If the controller ran between stages 1 and 2, then it will create an OSBuild instance and then run again after stage 2 and create another OSBuild instance.
If it ran after stage 2, it will create an OSBuild instance from the top of the configuration.

I don't see a scenario in which the instance does not exist. If we get that, I think we're in an error state. Please keep in mind that OSBuild instances are not deleted.

controllers/osbuildconfig_controller_test.go
ygalblum: It does, it updates the annotation

ygalblum (changes requested):

controllers/osbuildconfig_controller_test.go
bardielle: but the user configuration doesn't contain the annotation only the template and customization

controllers/osbuildconfig_controller.go
bardielle: So what we should do in that scenario? requeue for short duration? or Done?

controllers/osbuildconfig_controller_test.go
ygalblum: When I said both, I meant both `LastKnownUserConfiguration is different` and `webHookAnnotationKey has changed`

controllers/osbuildconfig_controller.go
ygalblum: I think we should treat it like any other (i.e. not check the error type)

api/v1alpha1/osbuild_types.go
```diff
@@ -46,37 +46,28 @@ type TriggeredBy string
...
+	AccessUrl string `json:"accessUrl,omitempty"`
 
 	// +optional
 	// ComposerIso is the URL for the iso that composer build returns before
 	// packaing with the kickstart
 	ComposerIso string `json:"composer_iso,omitempty"`
 }
 
-type OSBuildCondition struct {
+type Condition struct {
```
ygalblum, line 68: On one hand you generalized the name of the struct (same for `ConditionType`), but the description is still specific to OSBuild. Since we need to maintain a similar struct for the `OSBuildEnvConfig` and it seems like it is going to have the same structure, I think we need to generalize the descriptions as well

api/v1alpha1/osbuild_types.go
```diff
@@ -90,7 +81,17 @@ type OSBuildCondition struct {
...
 
-type OSBuildConditionType string
+type ConditionType string
+
+// These are the valid OSBuild condition types
+const (
+	// Whether OSBuild build is done and ready
+	ConditionReady ConditionType = "Ready"
+	// Whether OSBuild build is in progress
+	ConditionProgressing ConditionType = "Progressing"
```
ygalblum, line 91: Please change `Progressing` to `InProgress` (both const name and content)

controllers/osbuildconfig_controller.go
bardielle: Done

api/v1alpha1/osbuild_types.go
bardielle: Done

api/v1alpha1/osbuild_types.go
bardielle: Done

api/v1alpha1/osbuild_types.go
```diff
@@ -90,7 +81,17 @@ type OSBuildCondition struct {
 	LastTransitionTime *metav1.Time `json:"lastTransitionTime,omitempty" description:"last time the condition transit from one status to another"`
 }
 
-type OSBuildConditionType string
+type ConditionType string
+
+// These are the valid OSBuild condition types
+const (
+	// Whether OSBuild build is done and ready
```
ygalblum, line 88: The name of the const is general, so its documentation should also be general (so that we can use it in other controllers, e.g. OSBuildEnvConfig). Something like `The resource is ready`, `The resource failed`, `The resource is in progress`. 
We could also change the name and make it specific to OSBuild, but then we will need to duplicate the values

controllers/osbuild_controller.go
```diff
@@ -107,60 +100,45 @@ func (r *OSBuildReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 
-	if osBuild.Spec.Details != nil {
-		// build edge-container image
-		return r.buildEdgeContainerImage(ctx, logger, osBuild)
-	} else if osBuild.Spec.EdgeInstallerDetails != nil {
-		return r.buildEdgeInstallerImage(ctx, logger, osBuild)
+	if osBuild.Status.ComposeId == EmptyComposeID {
+		// if the image wasn't created yet - schedule a new build
+		logger.Info("create a new image")
+		// TODO: if target image type is edge-installer call postComposeEdgeInstallerImage
```
ygalblum, line 106: Just want to make sure. This doesn't create a loop of builds if I try to build an installer, right? (causing the OSBuildConfig controller to keep scheduling it)

ygalblum (changes requested):

controllers/osbuildconfig_controller.go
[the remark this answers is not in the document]
bardielle: I can understand it, It's really a rare condition 
I treated it anyway and added a UT

controllers/osbuild_controller.go
bardielle: Right - this is a todo comment for checking if the post compose request fo edge-installer is different and might be done in a different method

bardielle: Please mark it as done if it's fine by you

controllers/osbuildconfig_controller_test.go
bardielle: I updated it

controllers/osbuildconfig_controller.go
```diff
@@ -70,15 +83,160 @@ func (r *OSBuildConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		return r.updateOSBuildConfigCurrentStatus(ctx, logger, osBuildConfig)
+	}
+}
+
+func (r *OSBuildConfigReconciler) checkIfNewOSBuildInstanceIsNeeded(ctx context.Context, osBuildConfig *osbuilderv1alpha1.OSBuildConfig, logger logr.Logger) (bool, error) {
+	userConfiguration := r.getSortedUserConfiguration(osBuildConfig)
+	isNewOSBuildInstanceNeeded := false
+
+	if osBuildConfig.Status.LastKnownUserConfiguration == nil || !cmp.Equal(osBuildConfig.Status.LastKnownUserConfiguration, &userConfiguration) {
+		patch := client.MergeFrom(osBuildConfig.DeepCopy())
```
ygalblum, line 109: You should patch only once in this method, not separately for each check (this is also what fails the UT)

ygalblum (changes requested):

controllers/osbuildconfig_controller.go
```diff
@@ -70,15 +83,152 @@ func (r *OSBuildConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	userConfiguration := r.getSortedUserConfiguration(osBuildConfig)
+	isNewOSBuildInstanceNeeded := false
+	patch := client.MergeFrom(osBuildConfig.DeepCopy())
+
+	if osBuildConfig.Status.LastKnownUserConfiguration == nil || !cmp.Equal(osBuildConfig.Status.LastKnownUserConfiguration, &userConfiguration) {
+		osBuildConfig.Status.LastKnownUserConfiguration = &userConfiguration
+		isNewOSBuildInstanceNeeded = true
+	}
+
+	if osBuildConfig.Annotations == nil {
```
ygalblum, line 114: You can't return in this block because you are not updating the resource (if it changed).

ygalblum (changes requested):

controllers/osbuildconfig_controller.go
bardielle: I think you didn't refresh - this was a commit that I pushed before I saw your comment (to patch once for annotation and user configuration)

ygalblum: No, it's still here. If the configuration is different, but Annotations is nil you will return inside this block without updating the resource. I guess that in the UT Annotations is not nil.
BTW can it be nil? If not, than maybe this `if` is not needed

ygalblum (approved):

PR #189, opened 2022-09-07
supporting target image type qcow2

bardielle:


controllers/osbuild_controller.go
```diff
@@ -380,12 +381,14 @@ func (r *OSBuildReconciler) createImageRequest(osBuild *osbuildv1alpha1.OSBuild,
 	return &imageRequest, nil
 }
 
-func (r *OSBuildReconciler) getUploadOptions(osBuild *osbuildv1alpha1.OSBuild, targetImageType osbuildv1alpha1.TargetImageType) (*composer.UploadOptions, error) {
+func (r *OSBuildReconciler) getUploadOptions(osBuild *osbuildv1alpha1.OSBuild, targetImageType osbuildv1alpha1.TargetImageType, logger logr.Logger) (*composer.UploadOptions, error) {
 	var uploadOptions composer.UploadOptions
 	switch uploadTypeForTargetImageType[targetImageType] {
 	case composer.UploadTypesAwsS3:
+		logger.Info("upload to S3")
```
ygalblum, line 388: Are these logs really add anything? Whoever reads the logs already knows which images go to which target

bardielle: I will remove it...

ygalblum (commented):
You duplicated many of the tests (by adding an Entry per image type), but changing the parameter does not change anything in the code and the difference between them is not tested after the call to reconcile.
So, what's the benefit of having twice the amount of tests?

bardielle:
@ygal - I updated the relevant test to use DoAndReturn and check the target image type

controllers/osbuild_controller_test.go
```diff
@@ -250,54 +250,80 @@ var _ = Describe("OSBuild Controller", func() {
...
 		})
 
-		It("should requeue for long duration if failed on postCompose with an error", func() {
+		DescribeTable("should requeue for long duration if failed on postCompose with an error", func(targetImageType osbuildv1alpha1.TargetImageType) {
 			// given
-			composerClient.EXPECT().PostComposeWithResponse(requestContext, gomock.Any()).Return(nil, errFailed)
+			osbuildInstance.Spec.Details.TargetImage.TargetImageType = targetImageType
+			composerClient.EXPECT().PostComposeWithResponse(requestContext, gomock.Any()).DoAndReturn(
+				func(ctx context.Context, body composer.PostComposeJSONRequestBody, reqEditors ...interface{}) (*composer.PostComposeResponse, error) {
+					Expect(body.ImageRequest.ImageType == composer.ImageTypes(targetImageType))
```
ygalblum, line 262: I know this works, but I think it'd be nicer to use the test syntax:
`Expect(body.ImageRequest.ImageType).To(Equal(composer.ImageTypes(targetImageType)))`
(in all `DoAndReturn` implementations)

bardielle: Done

ygalblum (approved):

